PR #1, opened 2020-08-05
Add sshutil package

maraino:
### Description

This PR adds the implementation of ssh templates using the package sshutil. It also includes the randutil package from smallstep/cli as sshutil depends on it.

### DIFF
--- a/sshutil/options.go
+++ b/sshutil/options.go
@@ -1,0 +1,90 @@
+package sshutil
+
+import (
+	"bytes"
+	"encoding/base64"
+	"io/ioutil"
+	"text/template"
+
+	"github.com/Masterminds/sprig/v3"
+	"github.com/pkg/errors"
+	"go.step.sm/crypto/internal/step"
+)
+
+func getFuncMap(failMessage *string) template.FuncMap {
+	m := sprig.TxtFuncMap()
+	m["fail"] = func(msg string) (string, error) {
+		*failMessage = msg
+		return "", errors.New(msg)
+	}
+	return m
+}
+
+// Options are the options that can be passed to NewCertificate.
+type Options struct {
+	CertBuffer *bytes.Buffer
+}
+
+func (o *Options) apply(cr CertificateRequest, opts []Option) (*Options, error) {
+	for _, fn := range opts {
+		if err := fn(cr, o); err != nil {
+			return o, err
+		}
+	}
+	return o, nil
+}
+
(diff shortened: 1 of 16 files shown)


sshutil/templates.go
```diff
@@ -0,0 +1,195 @@
...
+
+// TemplateData is an alias for map[string]interface{}. It represents the data
+// passed to the templates.
+type TemplateData map[string]interface{}
+
+// CreateTemplateData returns a TemplateData with the given certificate type,
+// key id, principals, and the default extensions.
+func CreateTemplateData(ct CertType, keyID string, principals []string) TemplateData {
+	return TemplateData{
+		TypeKey:       ct.String(),
```
dopey, line 36: I didn't know that you could use variable names as keys when defining a map this way. cool.

sshutil/templates.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// SetCertificateRequest sets the simulated ssh certificate request the insecure
+// template data.
+func (t TemplateData) SetCertificateRequest(cr CertificateRequest) {
+	t.SetInsecure(CertificateRequestKey, cr)
+}
+
+// DefaultCertificate is the default template for an SSH certificate.
+const DefaultCertificate = `{
```
dopey, line 144: shouldn't this name be DefaultTemplate? otherwise the name will be sshutil.DefaultCertificate, which sort of implies  something different.

sshutil/options.go
```diff
@@ -0,0 +1,90 @@
...
+
+		tmpl, err := template.New("template").Funcs(funcMap).Parse(text)
+		if err != nil {
+			return errors.Wrapf(err, "error parsing template")
+		}
+
+		buf := new(bytes.Buffer)
+		data.SetCertificateRequest(cr)
+		if err := tmpl.Execute(buf, data); err != nil {
+			if terr.Message != "" {
```
dopey, line 55: Can you explain why we do this trickery with the error (rather than just return the error from Execute). I presume the error returned by execute has very little information.

dopey: And by explain, I just mean 'leave a comment'.

maraino: Sure, the error is very obscured, by text/template, so I just want to return the string in the template.

sshutil/templates.go
maraino: Yes, in fact in x509util use use things like `DefaultLeafTemplate`

PR #9, opened 2020-09-01
Do not enforce signature algorithm.

maraino:
### Description

This PR ignores the SignatureAlgorithm when no templates are used or in the template variable `.Insecure.CR`.

If SignatureAlgorithm is set in those cases, x509.CreateCertificate will fail if the signer cannot sign with that algorithm, e.g: RSA signer trying to sign ECDSA-SHA256.

This PR fixes some problematic cases with https://github.com/smallstep/crypto/pull/7

### DIFF
--- a/x509util/certificate.go
+++ b/x509util/certificate.go
@@ -53,12 +53,9 @@ func NewCertificate(cr *x509.CertificateRequest, opts ...Option) (*Certificate, 
 	}
 
 	// If no template use only the certificate request with the default leaf key
-	// usages. And do not enforce signature algorithm from the CSR, it might not
-	// be compatible with the certificate signer.
+	// usages.
 	if o.CertBuffer == nil {
-		cert := newCertificateRequest(cr).GetLeafCertificate()
-		cert.SignatureAlgorithm = 0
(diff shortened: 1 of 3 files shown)


dopey:
So, to clarify, the sig alg check can only fail in createCertificate if a template with a sigalg setting is used. Because the sigalg in the CSR will always be ignored. yes?

maraino:
> So, to clarify, the sig alg check can only fail in createCertificate if a template with a sigalg setting is used. Because the sigalg in the CSR will always be ignored. yes?

Yes, but it has to be a `Certificate` template, the problem was that only the CR available as a variable was used to create the `x509.Certificate`, and that variable was specifying a `SignatureAlgorithm`. 

If you create a Certificate template with an algorithm that is not supported by your signer, the problem is your template. We can also generate a template that takes everything from the CR ignoring the SignatureAlgorithm, but it won't be as easy as `{{ toJson .Insecure.CR }}`.

A `CertificateRequest` template can also fail to create a `x509.CertificateRequest` if you define a SignatureAlgorithm that is not supported by the signer. And again the problem is your template.

PR #10, opened 2020-09-21
Create a new default template for admin provisioners.

maraino:
### Description

This PR adds a new template were only the SANs and subject is read from the CR. The key usage and extended key usages are set in the template.

This template would become the default in k8ssa and admin-OIDC provisioners instead of using `{{ toJson .Insecure.CR }}`


dopey (approved):
lgtm

PR #15, opened 2020-12-30
tlsutil package

maraino:
### Description

This PR adds a new package including a refined version of the ca.Renewer in https://github.com/smallstep/certificates, as well as a ServerCredentials type that is a renewer for servers. The main difference is that the renew function can return different credentials depending on the SNI.


dopey (approved):
Looks like a lack of error case testing in server_credentials.go.

If you're comfortable with it, then looks good tome.

PR #16, opened 2021-03-30
randutil: add Alphabet

simaotwx:
Generates a string out of A-Za-z

Test: go test './...'
      make test


maraino:
@simaotwx Thanks for the contribution, I'll create a new tag with this.

maraino:
Here it is v0.8.1

simaotwx:
You're welcome and thanks for merging the change

PR #17, opened 2021-05-13
Add to randutil a method that returns a random uuid

maraino:
### Description

This PR adds `randutil.UUIDv4()` a new method that returns a version 4 uuid (random).


dopey (approved):
lgtm

PR #18, opened 2021-06-30
ValidateX5 should take a certificate chain rather

dopey:
- we can use these methods with in memory certificates


maraino (approved):
Looks good, but tag this as 0.9.0 because the API has changed

PR #22, opened 2021-09-15
Create method NewServerCredentialsFromFile

maraino:
### Description

This PR adds the new method `tlsutil.NewServerCredentialsFromFile()` that loads a certificate and its key from a file, and uses the standard renewer to renew it from disk. It also add `ServerCredentials.TLSConfig()` that returns a `*tls.Config` with the `GetCertificate` and `GetConfigForClient` set.

This PR also upgrade its dependencies, the new version of `gopkg.in/square/go-jose.v2` fixes a bug that was causing wrong thumbprints of Ed25519 keys. See square/go-jose@a10ff54


tlsutil/server_credentials.go
```diff
@@ -24,6 +27,36 @@ func NewServerCredentials(fn ServerRenewFunc) (*ServerCredentials, error) {
 	}, nil
 }
 
+// NewServerCredentialsFromFile returns a ServerCredentials that renews the
+// certificate from a file in disk.
```
dopey, line 31: file on disk

dopey (approved):
lgtm

tlsutil/server_credentials.go
maraino: Fixed with 746d3c4

PR #27, opened 2021-11-02
Initial support for additional SAN types

unreality:
This pull fixes [#431]( https://github.com/smallstep/certificates/issues/431#issue-752838546) and adds support for the following SAN types from [RFC5280](https://datatracker.ietf.org/doc/html/rfc5280.html#section-4.2.1.6):
* RegisteredID
* OtherName

It adds a check on certificates to see whether SAN types unsupported by `crypto/x509` are present, and if they are, uses `createSubjectAltNameExtension` to construct a SAN extension block, preventing the golang library from creating that extension. If the subjectAltName OID (2.5.29.17) is already present in the Certificate extensions list, no SAN extension is generated.

`SubjectAlternativeName.Set()` usage is avoided if unsupported SANs are present, so additional types were not added to the function.

For OtherName SANs, the type is assumed to be ASN1 `printable` unless specified in the Value of the SAN, eg:
```
{"type": "1.3.6.1.4.1.311.20.2.3", "value":"utf8;joe.user@example.com"}
```
will encode the OtherName value as a UTF8 string instead of printable. The following types are supported based on [encoding/asn1](https://pkg.go.dev/encoding/asn1) :
* `int` (ASN Integer)
* `oid` (OID)
* `utf8` (UTF8 String)
* `printable` (PrintableString)
* `numeric` (NumericString)
* `ia5` (IA5String)
Also available is a special type:
* `raw` - Accepts a base64 encoded value which is marshalled directly into the certificate. This allows users to manually craft an ASN1 data structure and add it into the template manually in order to support types like BMPString, HEX values, and other types that `encoding/asn1` does not support.

x509 templates can be used to specify the additional SAN types, eg:

```
{
  "subject": {{ toJson .Subject }},
  "sans": [
    {"type": "email", "value":"rfc822@example.com"},
    {"type": "dns", "value": "a.dns.example.com"},
    {"type": "dns", "value": "b.dns.example.com"},
    {"type": "dns", "value": "bc.dns.example.com"},
    {"type": "ip", "value": "10.10.10.10"},
    {"type": "registeredID", "value": "1.3.6.1.4.1.311.20.2.3"},
    {"type": "1.3.6.1.4.1.311.20.2.3", "value":"utf8;joe.user@example.com"}
  ],
{{- if typeIs "*rsa.PublicKey" .Insecure.CR.PublicKey }}
  "keyUsage": ["keyEncipherment", "digitalSignature"],
{{- else }}
  "keyUsage": ["digitalSignature"],
{{- end }}
  "extKeyUsage": ["serverAuth", "clientAuth"]
}
```

Stub code is added for `X400Address`, `DirectoryName`, and `EDIPartyName` types, but they currently return not implemented errors if they are used.


x509util/extensions.go
```diff
@@ -500,3 +533,170 @@ func (s *SerialNumber) UnmarshalJSON(data []byte) error {
...
+
+	// Now iterate over all the SANs and construct the raw ASN1 encoded values and place them in rawValues
+	for _, san := range allSANs {
+
+		switch san.Type {
+
+		case EmailType:
+			rawValues = append(rawValues, asn1.RawValue{Tag: nameTypeEmail, Class: 2, Bytes: []byte(san.Value)})
+		case DNSType:
+			rawValues = append(rawValues, asn1.RawValue{Tag: nameTypeDNS, Class: 2, Bytes: []byte(san.Value)})
```
maraino, line 590: Email and DNS should be encoded in IA5String (ASCII) according to RFC5280. The current behavior is to fail. But alternatively, we can use `tlsutils.SanitizeName`.

x509util/extensions.go
```diff
@@ -500,3 +533,170 @@ func (s *SerialNumber) UnmarshalJSON(data []byte) error {
...
+		case X400AddressType:
+			fallthrough
+		case DirectoryNameType:
+			fallthrough
+		case EDIPartyNameType:
+			return nil, fmt.Errorf("unimplemented SAN type %s", san.Type)
+		case URIType:
+			rawValues = append(rawValues, asn1.RawValue{Tag: nameTypeURI, Class: 2, Bytes: []byte(san.Value)})
```
maraino, line 598: URI is also limited to IA5String

x509util/certificate.go
```diff
@@ -79,15 +80,25 @@ func (c *Certificate) GetCertificate() *x509.Certificate {
 	cert.PublicKey = c.PublicKey
 	cert.PublicKeyAlgorithm = c.PublicKeyAlgorithm
 
-	// SANs are directly converted.
-	cert.DNSNames = c.DNSNames
-	cert.EmailAddresses = c.EmailAddresses
-	cert.IPAddresses = c.IPAddresses
-	cert.URIs = c.URIs
+	if c.hasExtendedSANs() && !c.hasExtension(subjectAlternativeNameOID) {
```
maraino, line 83: Nice, only use createSubjectAltNameExtension if it has SANs non supported by Go.

x509util/extensions.go
```diff
@@ -500,3 +533,170 @@ func (s *SerialNumber) UnmarshalJSON(data []byte) error {
...
+			}
+		case RegisteredIDType:
+			if san.Value != "" {
+				oid, err := parseObjectIdentifier(san.Value)
+				if err == nil {
+					rawBytes, _ := asn1.MarshalWithParams(oid, fmt.Sprintf("tag:%d", nameTypeRegisteredID))
+					rawValues = append(rawValues, asn1.RawValue{FullBytes: rawBytes})
+				}
+			}
+		default:
```
maraino, line 615: We have also the AutoType and "" that is equivalent to AutoType, see `SubjectAlternativeName.Set()`.

x509util/extensions.go
```diff
@@ -500,3 +533,170 @@ func (s *SerialNumber) UnmarshalJSON(data []byte) error {
...
+						// marshaller. By using this type users can add ASN1 datatypes manually into templates
+						// to support some unsupported types like BMPString or HEX types
+						rawBytes, err = base64.StdEncoding.DecodeString(sanValue)
+					case "utf8":
+						fallthrough
+					case "ia5":
+						fallthrough
+					case "numeric":
+						fallthrough
+					case "printable":
```
maraino, line 663:
```suggestion
					case "utf8", "ia5", "numeric", "printable":
```

maraino: I guess `MarshalWithParams` will fail on ia5 and numeric if the sanValue is not valid. We should add tests for this, and for this method.

x509util/extensions.go
```diff
@@ -500,3 +533,170 @@ func (s *SerialNumber) UnmarshalJSON(data []byte) error {
...
+
+					// OtherName SANs are an ASN1 sequence containing OID and Value
+					otherName := OtherName{
+						OID:   oid,
+						Value: OtherNameValue{V: otherNameValue},
+					}
+					generalNameBytes, _ := asn1.MarshalWithParams(otherName, fmt.Sprintf("tag:%d", nameTypeOtherName))
+					rawValues = append(rawValues, asn1.RawValue{FullBytes: generalNameBytes})
+				} else {
+					return nil, fmt.Errorf("unsupported SAN type %s", san.Type)
```
maraino, line 684: We should return errors from internal methods, if not we lose track of the stack trace.

x509util/extensions.go
```diff
@@ -500,3 +533,170 @@ func (s *SerialNumber) UnmarshalJSON(data []byte) error {
...
+				}
+			}
+
+		}
+
+	}
+
+	// Now marshal the rawValues into the ASN1 sequence, and create an Extension object to hold the extension
+	rawBytes, _ := asn1.Marshal(rawValues)
```
maraino, line 693: Do not ignore errors, and use errors.Wrap(err, "error marshaling to asn1") to add a stack trace.

x509util/extensions.go
```diff
@@ -500,3 +533,170 @@ func (s *SerialNumber) UnmarshalJSON(data []byte) error {
...
+						fallthrough
+					case "printable":
+						rawBytes, err = asn1.MarshalWithParams(sanValue, valueType)
+					default:
+						// if it's an unknown type, default to printable - but use the entire value specified in case there is a semicolon in the value
+						rawBytes, err = asn1.MarshalWithParams(san.Value, "printable")
+					}
+
+					if err != nil {
+						return nil, fmt.Errorf("could not marshal ASN1 values: %v", err)
```
maraino, line 671: Use `errors.Wrapf(err, "error marshaling asn1")`. In fact, add this after each err so we can get the exact line of an error.

x509util/extensions.go
```diff
@@ -500,3 +533,170 @@ func (s *SerialNumber) UnmarshalJSON(data []byte) error {
...
+		}
+
+	}
+
+	// Now marshal the rawValues into the ASN1 sequence, and create an Extension object to hold the extension
+	rawBytes, _ := asn1.Marshal(rawValues)
+
+	subjectAltNameExtension := Extension{
+		ID:       subjectAlternativeNameOID,
+		Critical: false, // TODO this should be true if Certificate Subject is blank
```
maraino, line 697: You need to add a parameter to this function to know if the subject is empty.

x509util/extensions.go
maraino: Use errors.Wrap in all previous errors if they are not coming from a method from this library.

maraino (changes requested):

x509util/extensions.go
```diff
@@ -158,6 +191,160 @@ func (s SubjectAlternativeName) Set(c *x509.Certificate) {
 	}
 }
 
+func (s SubjectAlternativeName) RawValue() (*asn1.RawValue, error) {
+	var asnValue asn1.RawValue
+
+	switch s.Type {
+	case "", AutoType:
+		// autotype requires us to find out what the type is
+		if _, err := parseObjectIdentifier(s.Value); err != nil {
```
maraino, line 200: An IP will be parsed as an OID, we cannot parse this type automatically.

maraino (changes requested):
Hi @unreality, I've found a couple of problems, but the rest I believe is ok.

maraino:
Can you fix the linter errors? And it would be nice to have tests too. The coverage is of this package is quite high, and with this is going to go down:
```
$ go test -cover ./x509util/...
ok  	go.step.sm/crypto/x509util	1.023s	coverage: 98.8% of statements
```

unreality:
I'll have a look, from what i saw some of the linter errors werent in my additions? I'll have to read up about golang tests since ive never written them before, so it might take a while to get those done :)

maraino:
> I'll have a look, from what i saw some of the linter errors werent in my additions? I'll have to read up about golang tests since ive never written them before, so it might take a while to get those done :

I can merge first, but make sure to fix the errors. One of them is that you're doing a import cycle probably because tlsutils depends on x509utils, and the other way around, it might be because of the import of tlsutils.SanitizeName. We can move it to x509util and add a pointer to it in tlsutil to keep backward compatibility:

```
var SanitizeName = x509util.SanitizeName
```

burnbabyburn:
thx for working/contributing on this!!! looking forward to see the issues resolved 👍

unreality:
I've updated this branch to be in line with master, @maraino could you review?

Im looking at tests at the moment, however the existing tests largely use the golang X509 libraries to verify certificate contents etc. Given we are adding Extended SANs that the X509 libs cant process, do you have any good ideas on how to verify the returned data?

maraino:
Hi @unreality, thanks for updating this, I won't merge it right away, because I'm gonna be working on device attestation SANs, and I might need to change things. It might take a few weeks, I'm going on vacations :)

x509util/extensions.go
```diff
@@ -499,3 +683,93 @@ func (s *SerialNumber) UnmarshalJSON(data []byte) error {
...
+		}.RawValue()
+
+		if err != nil {
+			return nil, err
+		}
+
+		rawValues = append(rawValues, *rawValue)
+	}
+
+	for _, emailAddress := range c.DNSNames {
```
maraino, line 713: This should be c.EmailAddresses, but I will fix it as I'm doing other changes.

x509util/extensions.go
```diff
@@ -157,6 +189,158 @@ func (s SubjectAlternativeName) Set(c *x509.Certificate) {
...
+			// The default type is printable, but if the value is prefixed with a type, use that
+			var valueType = "printable"
+			var sanValue = s.Value
+			var rawBytes []byte
+			var err error
+
+			if strings.Contains(s.Value, sanTypeSeparator) {
+				valueType = strings.Split(s.Value, sanTypeSeparator)[0]
+				sanValue = s.Value[len(valueType)+1:]
+			}
```
maraino, line 269: @unreality is there any reason to use `;` instead of `:` as the separator? You're linking there to the OpenSSL docs that use `[modifier,]type[:value]`.

maraino (commented):
I'm making some PR changes; instead of fixing them, just take a look at my questions.

x509util/extensions.go
unreality: Not that I can recall off the top of my head...colon is probably fine to use

x509util/extensions.go
```diff
@@ -51,13 +52,44 @@ var (
...
+	nameTypeDNS       = 2
+	//nameTypeX400         = 3
+	//nameTypeDirectory    = 4
+	//nameTypeEDI          = 5
+	nameTypeURI          = 6
+	nameTypeIP           = 7
+	nameTypeRegisteredID = 8
 )
 
+const sanTypeSeparator = ";"
```
maraino, line 79: I'm changing `sanTypeSeparator` to `:`.

PR #33, opened 2022-01-27
Use RSA-SHA2-256 when signing ssh certificates with an RSA key

maraino:
### Description

This PR replaces the default RSA hash algorithm of x/crypto/ssh, SHA-1 (ssh-rsa) with SHA256 (rsa-sha2-256).

This new signature scheme has been supported since 2016 with OpenSSH 7.2. And on [OpenSSH 8.2](https://www.openssh.com/txt/release-8.2) was announced that ssh-rsa will be deprecated, and it has been removed in [OpenSSH 8.8](https://www.openssh.com/txt/release-8.8)


dopey (approved):
lgtm

PR #34, opened 2022-02-01
Use password prompt if available for open ssh private key

dopey:


maraino (approved):
looks good, but I think it makes sense to replace it with
https://github.com/smallstep/crypto/blob/6c198f63d29d7d53aaf071eae60247369ab703ac/pemutil/pem.go#L78-L89

See https://github.com/smallstep/crypto/blob/6c198f63d29d7d53aaf071eae60247369ab703ac/pemutil/pem.go#L388-L398

maraino (approved):
👍

PR #36, opened 2022-02-28
Add Bytes() function

hslatman:
We could call `Salt()` or the other way around, but it's only a little bit of duplicate code, so I figured it's OK to do like this.


maraino (approved):
lgtm

PR #37, opened 2022-03-11
Make it easier to sign a JWT with a key.

maraino:
### Description

With this change, the signature algorithm will be guessed if not provided.


dopey (approved):
lgtm

PR #40, opened 2022-03-15
MiniCA

maraino:
### Description

This PR implements a mini certificate authority. 

When we're doing unit tests, having a simple CA to play with is generally useful. This new package adds a simple implementation for a CA.

@azazeal: This is something that I wanted to do for some time, and today had some time.
I will create something like [this](https://github.com/jsha/minica) using this new package.

### DIFF
--- a/minica/minica.go
+++ b/minica/minica.go
@@ -11,8 +11,8 @@ import (
 	"golang.org/x/crypto/ssh"
 )
 
-// MiniCA is the implementation of a simple X.509 and SSH CA.
-type MiniCA struct {
+// CA is the implementation of a simple X.509 and SSH CA.
+type CA struct {
 	Root          *x509.Certificate
 	Intermediate  *x509.Certificate
 	Signer        crypto.Signer
@@ -22,7 +22,7 @@ type MiniCA struct {
 
 // New creates a new MiniCA, the custom options allows to overwrite templates,
 // signer types and certificate names.
-func New(opts ...Option) (*MiniCA, error) {
+func New(opts ...Option) (*CA, error) {
 	now := time.Now()
 	o := newOptions().apply(opts)
 
@@ -90,7 +90,7 @@ func New(opts ...Option) (*MiniCA, error) {
 		return nil, err
 	}
 
-	return &MiniCA{
+	return &CA{
 		Root:          root,
 		Intermediate:  intermediate,
 		Signer:        intSigner,
@@ -100,7 +100,7 @@ func New(opts ...Option) (*MiniCA, error) {
 }
 
 // Sign signs an X.509 certificate template using the intermediate certificate.
-func (c *MiniCA) Sign(template *x509.Certificate) (*x509.Certificate, error) {
+func (c *CA) Sign(template *x509.Certificate) (*x509.Certificate, error) {
 	if template.NotBefore.IsZero() {
 		template.NotBefore = time.Now()
 	}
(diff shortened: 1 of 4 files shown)


minica/minica.go
```diff
@@ -0,0 +1,158 @@
...
+		Signer:        intSigner,
+		SSHHostSigner: sshHostSigner,
+		SSHUserSigner: sshUserSigner,
+	}, nil
+}
+
+// Sign signs an X.509 certificate template using the intermediate certificate.
+func (c *MiniCA) Sign(template *x509.Certificate) (*x509.Certificate, error) {
+	if template.NotBefore.IsZero() {
+		template.NotBefore = time.Now()
```
azazeal, line 105: I think since we may end up mutating the input we should document the function to include this case?

minica/minica.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+
+	return c.Sign(cert)
+}
+
+// SignSSH signs an SSH host or user certificate.
+func (c *MiniCA) SignSSH(cert *ssh.Certificate) (*ssh.Certificate, error) {
+	if cert.ValidAfter == 0 {
+		cert.ValidAfter = uint64(time.Now().Unix())
```
azazeal, line 143: this is also mutating the input - not sure if we can avoid this :/

azazeal (commented):
Few comments I came up with after a quick read. Nice work!

minica/minica.go
```diff
@@ -0,0 +1,158 @@
...
+	"time"
+
+	"go.step.sm/crypto/sshutil"
+	"go.step.sm/crypto/x509util"
+	"golang.org/x/crypto/ssh"
+)
+
+// MiniCA is the implementation of a simple X.509 and SSH CA.
+type MiniCA struct {
+	Root          *x509.Certificate
```
azazeal, line 16: Not quite sure if a CA should be immutable after initialization. Maybe not export these fields directly and instead export getters in case we do want for people to be accessing them?

minica/minica.go
```diff
@@ -0,0 +1,158 @@
...
+	"fmt"
+	"time"
+
+	"go.step.sm/crypto/sshutil"
+	"go.step.sm/crypto/x509util"
+	"golang.org/x/crypto/ssh"
+)
+
+// MiniCA is the implementation of a simple X.509 and SSH CA.
+type MiniCA struct {
```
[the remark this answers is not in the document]
maraino: Fixed with 99fbede

minica/minica.go
maraino: Fixed with cd2e52d

minica/minica.go
maraino: I think making these fields public adds some extra flexibility to the package without requiring some extra options. The user of this package can take precautions to avoid any problem.

minica/minica.go
maraino: Fixed with https://github.com/smallstep/crypto/commit/cd2e52da0b66ce3a0fde8470ba7751d1e9a970eb

minica/minica_test.go
```diff
@@ -0,0 +1,541 @@
...
+			return reflect.DeepEqual(auth, ca.SSHHostSigner.PublicKey())
+		},
+	}
+	if err := checker.CheckHostKey("ssh.test.com:22", &net.IPAddr{IP: net.IP{1, 2, 3, 4}}, got); err != nil {
+		t.Errorf("CertChecker.CheckHostKey() error = %v", err)
+	}
+
+	// Validate mutation
+	if template.ValidAfter != 0 {
+		t.Errorf("CA.SignSSH() mutated template.ValidAfter")
```
azazeal, line 474: there's a lot of `Errorf` calls in which you don't interpolate anything, I feel as if `t.Error` would be more appropriate in these cases.

minica/minica_test.go
```diff
@@ -0,0 +1,541 @@
...
+					}
+				default:
+					t.Fatalf("unknown cert type %v", tt.wantCertType)
+				}
+			}
+		})
+	}
+}
+
+func TestCA_SignSSH_mutation(t *testing.T) {
```
azazeal, line 439: I see a naming pattern with underscores for tests. Not quite sure why the linter isn't complaining but maybe `TestCASignSSHMutation` should be a more idiomatic name? This would apply to the rest as well.

maraino: A well-known tool named [gotests](https://github.com/cweill/gotests) defaults to `TestName` for functions and `TestStruct_Name` for methods. I'm adding another `_` when I want to add a test for a special use case, so it doesn't look like another method.

minica/minica_test.go
maraino: Fixed with 98fa65a

azazeal (approved):

PR #42, opened 2022-05-16
Populate ExtraNames from Names

maraino:
### Description

This commit will populate the subject ExtraNames in a certificate request from the information present in Names as long as it is not one of the default fields. This way, we can extract non-default fields as well as populate them in the final certificate if necessary.

It also uses the proper type for the deprecated email address attribute in the subject.

Related to smallstep/certificates#916

### DIFF
--- a/x509util/name.go
+++ b/x509util/name.go
@@ -53,7 +53,7 @@ func newName(n pkix.Name) Name {
 		PostalCode:         n.PostalCode,
 		SerialNumber:       n.SerialNumber,
 		CommonName:         n.CommonName,
-		ExtraNames:         newDistinguisedNames(n.Names),
+		ExtraNames:         newExtraNames(n.Names),
 	}
 }
 
@@ -149,9 +149,9 @@ type DistinguishedName struct {
 	Value interface{}      `json:"value"`
 }
 
-// newDistinguisedNames returns a list of DistinguishedName with the attributes not
+// newExtraNames returns a list of DistinguishedName with the attributes not
 // present in attributeTypeNames.
-func newDistinguisedNames(atvs []pkix.AttributeTypeAndValue) []DistinguishedName {
+func newExtraNames(atvs []pkix.AttributeTypeAndValue) []DistinguishedName {
 	var extraNames []DistinguishedName
 	for _, atv := range atvs {
 		if _, ok := attributeTypeNames[atv.Type.String()]; !ok {
--- a/x509util/name.go
+++ b/x509util/name.go
@@ -104,7 +104,7 @@ func (s Subject) Set(c *x509.Certificate) {
 		PostalCode:         s.PostalCode,
 		SerialNumber:       s.SerialNumber,
 		CommonName:         s.CommonName,
-		ExtraNames:         fromDistinguisedNames(s.ExtraNames),
+		ExtraNames:         fromDistinguishedNames(s.ExtraNames),
 	}
 }
 
@@ -138,7 +138,7 @@ func (i Issuer) Set(c *x509.Certificate) {
 		PostalCode:         i.PostalCode,
 		SerialNumber:       i.SerialNumber,
 		CommonName:         i.CommonName,
-		ExtraNames:         fromDistinguisedNames(i.ExtraNames),
+		ExtraNames:         fromDistinguishedNames(i.ExtraNames),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


x509util/name_test.go
```diff
@@ -27,8 +27,9 @@ func Test_newName(t *testing.T) {
 			PostalCode:         []string{"The postalCode"},
 			SerialNumber:       "The serialNumber",
 			CommonName:         "The commonName",
-			ExtraNames: []pkix.AttributeTypeAndValue{
-				{Type: asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 9, 1}, Value: "jane@example.com"},
+			Names: []pkix.AttributeTypeAndValue{
+				{Type: asn1.ObjectIdentifier{2, 5, 4, 3}, Value: "The commonName"},
```
hslatman, line 31: This instance of the `The commonName` is ignored, right? It may be clearer to use a different value than the one that's already set in `CommonName`, so that a change in the function will result in a different test outcome.

x509util/name.go
```diff
@@ -131,24 +149,44 @@ type DistinguishedName struct {
 	Value interface{}      `json:"value"`
 }
 
+// newDistinguisedNames returns a list of DistinguishedName with the attributes not
+// present in attributeTypeNames.
 func newDistinguisedNames(atvs []pkix.AttributeTypeAndValue) []DistinguishedName {
```
hslatman, line 154: Maybe the function name can reflect the fact that it excludes the attribute names? It might get overly long doing so though, like `newDistinguishedNamesExcludingExistingAttributes`. Don't have a better proposal atm 😅

x509util/name.go
```diff
@@ -35,7 +53,7 @@ func newName(n pkix.Name) Name {
 		PostalCode:         n.PostalCode,
 		SerialNumber:       n.SerialNumber,
 		CommonName:         n.CommonName,
-		ExtraNames:         newDistinguisedNames(n.ExtraNames),
+		ExtraNames:         newDistinguisedNames(n.Names),
```
hslatman, line 56: I think you kept the `ExtraNames` property for backwards compatibility? Should we maybe also provide the `Names` property, which could contain the same data?

x509util/name.go
```diff
@@ -9,6 +9,24 @@ import (
 	"github.com/pkg/errors"
 )
 
+// attributeTypeNames are the subject attributes managed by Go and this package.
+// newDistinguisedNames will populate .Insecure.CR.ExtraNames with the
+// attributes not present on this map.
```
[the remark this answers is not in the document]
maraino: It's a copy&paste from Go's source code.

x509util/name_test.go
maraino: I wanted to "imitate" a real case, Names will have all the attributes present in the subject C, O, OU, CN, ... I will add all of them to make sure the map is correct.

maraino: Fixed with 4434f23

x509util/name.go
maraino: Fixed with c93d3e3 (`newExtraNames`)

x509util/name.go
```diff
@@ -35,7 +53,7 @@ func newName(n pkix.Name) Name {
 		PostalCode:         n.PostalCode,
 		SerialNumber:       n.SerialNumber,
 		CommonName:         n.CommonName,
-		ExtraNames:         newDistinguisedNames(n.ExtraNames),
+		ExtraNames:         newExtraNames(n.Names),
```
hslatman, line 56: Much better name 😄

x509util/name.go
```diff
@@ -131,24 +149,44 @@ type DistinguishedName struct {
...
+			})
+		}
 	}
 	return extraNames
 }
 
+// fromDistinguisedNames converts a list of DistinguisedName to
+// []pkix.AttributeTypeAndValue. Note that this method has a special case to
+// encode the emailAddress deprecated field (1.2.840.113549.1.9.1).
 func fromDistinguisedNames(dns []DistinguishedName) []pkix.AttributeTypeAndValue {
```
hslatman, line 170: Needs update on the calling side too.

```suggestion
// fromDistinguishedNames converts a list of DistinguishedName to
// []pkix.AttributeTypeAndValue. Note that this method has a special case to
// encode the deprecated emailAddress field (1.2.840.113549.1.9.1).
func fromDistinguishedNames(dns []DistinguishedName) []pkix.AttributeTypeAndValue {
```

hslatman (approved):
lgtm

Just the suggestion and usage of the function to be fixed.

x509util/name.go
maraino: Fixed with 487ffed

PR #57, opened 2022-08-18
DirectoryName and HardwareModuleName SANs

### LINKED ISSUE #55: Add support for HardwareModule SAN
### Description

Add support for the hardware module SAN if it is finally defined in https://datatracker.ietf.org/doc/html/draft-bweeks-acme-device-attest-01#section-4

maraino:
### Description

This PR adds support for the DirectoryName and HardwareModuleName SANs. It also allows marshaling the PermanentIdentifier SAN with an assigner OID.

Fixes #55

### DIFF
--- a/x509util/extensions.go
+++ b/x509util/extensions.go
@@ -116,22 +116,20 @@ type otherName struct {
 //	  assigner           OBJECT IDENTIFIER OPTIONAL
 //	}
 type PermanentIdentifier struct {
-	IdentifierValue string                `asn1:"utf8,optional" json:"identifier"`
-	Assigner        asn1.ObjectIdentifier `asn1:"optional" json:"assigner,omitempty"`
+	Identifier string           `json:"identifier,omitempty"`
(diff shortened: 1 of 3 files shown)


x509util/extensions.go
```diff
@@ -63,20 +64,21 @@ const (
 	IPType                  = "ip"
 	RegisteredIDType        = "registeredID"
 	PermanentIdentifierType = "permanentIdentifier"
+	HardwareModuleType      = "hardwareModule"
```
hslatman, line 67: Not a blocker, but maybe this should be `HardwareModuleNameType` and `hardwareModuleName` to be consistent

hslatman (approved):

x509util/extensions.go
maraino: Yes, it should be.

maraino: Fixed with fd6d5e5

hslatman (approved):

PR #58, opened 2022-08-18
Add TemplateData.SetWebhooks method

areed:
### Description
Adds methods to sshutil.TemplateData and x509util.TemplateData to set arbitrary data under the key `Webhooks`.

💔Thank you!

### DIFF
--- a/x509util/templates.go
+++ b/x509util/templates.go
@@ -16,6 +16,7 @@ const (
 	CertificateRequestKey = "CR"
 	AuthorizationCrtKey   = "AuthorizationCrt"
 	AuthorizationChainKey = "AuthorizationChain"
+	WebhooksKey           = "Webhooks"
 )
 
 // TemplateError represents an error in a template produced by the fail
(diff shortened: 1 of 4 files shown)


x509util/templates.go
```diff
@@ -107,6 +108,17 @@ func (t TemplateData) SetCertificateRequest(cr *x509.CertificateRequest) {
 	t.SetInsecure(CertificateRequestKey, newCertificateRequest(cr))
 }
 
+// SetWebhooks sets the given webhook response in the webhooks template data.
+func (t TemplateData) SetWebhooks(webhookName string, data interface{}) {
```
maraino, line 112: It looks like it's assigning only one webhook:
```suggestion
// SetWebhook sets the given webhook response in the webhooks template data.
func (t TemplateData) SetWebhook(webhookName string, data interface{}) {
```

maraino: We also should add a method like this to `sshutil/templates.go`

areed: ade8a1f

maraino (approved):

PR #59, opened 2022-08-18
Require at least 2048 bits

maraino:
### Description

This PR requires generating at least 2048 bits RSA keys, although this value is configurable.


dopey (approved):
lgtm

PR #61, opened 2022-08-23
Insecure mode

maraino:
### Description

This PR adds a new method, `keyutil.Insecure()`, that allows the creation of insecure RSA keys. This can be required, for example, by `step` to create legacy keys.

Note: I'm not using atomic.Bool because it is not available in 1.18

### DIFF
--- a/keyutil/key.go
+++ b/keyutil/key.go
@@ -32,6 +32,16 @@ var (
 	MinRSAKeyBytes = 256
 )
 
+// Insecure removes minimum limits on on the RSA key size and returns a function
+// to revert the configuration.
+func Insecure() (revert func()) {
+	minRSAKeyBytes := MinRSAKeyBytes
+	MinRSAKeyBytes = 0
+	return func() {
+		MinRSAKeyBytes = minRSAKeyBytes
+	}
+}
+
 // PublicKey extracts a public key from a private key.
 func PublicKey(priv interface{}) (crypto.PublicKey, error) {
 	switch k := priv.(type) {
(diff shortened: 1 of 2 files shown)


keyutil/key_test.go
```diff
@@ -540,3 +540,41 @@ func TestVerifyPair(t *testing.T) {
 		})
 	}
 }
+
+func TestInsecure(t *testing.T) {
```
azazeal, line 544: I don't believe you gain anything by including "RSA" in your strings.

keyutil/key_test.go
```diff
@@ -540,3 +540,41 @@ func TestVerifyPair(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		run     func(t *testing.T) error
+		wantErr bool
+	}{
+		{"ok RSA 2048", func(t *testing.T) (err error) {
```
azazeal, line 550: I'd include a call to `t.Helper` in the runners of this table-driven test.

keyutil/key.go
```diff
@@ -32,6 +33,24 @@ var (
 	MinRSAKeyBytes = 256
 )
 
+type atomicBool int32
```
azazeal, line 36: I'd have gone with a simple variable (accessed atomically) but what you did offers a straighter line to a potential upgrade for 1.19.

azazeal (approved):
LGTM

keyutil/key_test.go
maraino: Because I'm not doing any `t.Error'-like method, there's no point in using `t.Helper()` to skip the function in the error information. I'm just passing `testing.T` to use the `Cleanup` method, I can totally avoid it with a defer too.

PR #62, opened 2022-08-23
Add `ValidateTemplate` function

hslatman:


maraino (commented):
I'm ok with `ValidateTemplateData`, but if we use `ValidateTemplate` to fail when creating a template, we don't give our users much flexibility. For example, if you run `fail` if something is not present, the `tmpl.Execute()` will fail, and the validation will fail.

I would remove `ValidateTemplate`

internal/templates/validate.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// resulting JSON should be valid; if not, the template was not formatted correctly
+	if ok := json.Valid(buf.Bytes()); !ok {
+		// determine what's wrong with the JSON exactly; the `Valid` method doesn't return that
+		var m map[string]interface{}
+		if err := json.NewDecoder(buf).Decode(&m); err != nil {
+			return fmt.Errorf("invalid JSON: %w", enrichJSONError(err))
+		}
+	}
+
```
maraino, line 41: After this morning discussions, perhaps the most sensible thing would be to just validate with the Parse.
```suggestion
```

hslatman:
Discussed this with @azazeal. We tried `Option("missingkey=zero"`, but that results in nil pointer errors. There may be ways to circumvent these, but not trivial to do. Also discussed trying to come up with a set of properties we know should work and providing these as template data for full execution and parsing. This is also not trivial to do, but might be worth it at a later time.

In the tests I have two test cases that won't return an error when validating the template, but would result in a failure when signing certs. It would be nice to negate those tests in the future with improved validation.

internal/templates/validate.go
```diff
@@ -0,0 +1,72 @@
...
+	"text/template"
+)
+
+// ValidateTemplate validates a text template results in valid JSON
+// when it's executed with empty template data. If template execution
+// results in invalid JSON, the template is invalid. When the template
+// is valid, it can be used safely. A valid template can still result
+// in invalid JSON when non-empty template data is provided.
+func ValidateTemplate(text string) error {
+
```
maraino, line 16:
```suggestion
```

maraino (approved):
Looks good but I will change the arguments from string to []byte before merging

PR #63, opened 2022-08-26
YubiKey Attestation

maraino:
### Description

This PR adds support for attestation using YubiKeys. It also adds the default template used with ACME device-attest-01.

### DIFF
--- a/kms/yubikey/yubikey.go
+++ b/kms/yubikey/yubikey.go
@@ -199,6 +199,11 @@ func (k *YubiKey) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, e
 }
 
 // CreateAttestation creates an attestation certificate from a YubiKey slot.
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
+// release.
 func (k *YubiKey) CreateAttestation(req *apiv1.CreateAttestationRequest) (*apiv1.CreateAttestationResponse, error) {
 	slot, err := getSlot(req.Name)
 	if err != nil {
--- a/kms/yubikey/yubikey.go
+++ b/kms/yubikey/yubikey.go
@@ -170,13 +170,21 @@ func (k *YubiKey) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, e
 		return nil, err
 	}
 
+	pin := k.pin
+	if pin == "" {
+		// Attempt to get the pin from the uri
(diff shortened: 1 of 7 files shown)


kms/yubikey/yubikey.go
```diff
@@ -190,6 +218,35 @@ func (k *YubiKey) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, e
...
+	if err != nil {
+		return nil, err
+	}
+
+	cert, err := k.yk.Attest(slot)
+	if err != nil {
+		return nil, errors.Wrap(err, "error attesting slot")
+	}
+
+	intermediate, err := k.yk.Certificate(slotAttestation)
```
hslatman, line 238: I guess this does the same as `k.yk.AttestationCertificate()`?

kms/yubikey/yubikey.go
```diff
@@ -190,6 +218,35 @@ func (k *YubiKey) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, e
...
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
+// release.
+func (k *YubiKey) CreateAttestation(req *apiv1.CreateAttestationRequest) (*apiv1.CreateAttestationResponse, error) {
+	slot, err := getSlot(req.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	cert, err := k.yk.Attest(slot)
```
hslatman, line 233: Callers of this function would call this after `k.yk.GenerateKey` with the name of the slot in which the key was created, correct?

hslatman (approved):

kms/yubikey/yubikey.go
maraino: Yes, I didn't realize that method was present.
```go
// AttestationCertificate returns the YubiKey's attestation certificate, which
// is unique to the key and signed by Yubico.
func (yk *YubiKey) AttestationCertificate() (*x509.Certificate, error) {
	return yk.Certificate(slotAttestation)
}
```

kms/yubikey/yubikey.go
maraino: For example, you can create a key using:

```console
$ step kms create yubikey:slot-id=85
-----BEGIN PUBLIC KEY-----
...
-----END PUBLIC KEY-----
$ step kms attest yubikey:slot-id=85
-----BEGIN CERTIFICATE-----
...
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
...
-----END CERTIFICATE-----
```

PR #64, opened 2022-09-09
Convert x509 to certificate

areed:
### Description
Add utility to convert an x509.Certificate to an x509util.Certificate.

💔Thank you!

### DIFF
--- a/x509util/certificate.go
+++ b/x509util/certificate.go
@@ -55,7 +55,7 @@ func NewCertificate(cr *x509.CertificateRequest, opts ...Option) (*Certificate, 
 	// If no template use only the certificate request with the default leaf key
 	// usages.
 	if o.CertBuffer == nil {
-		return newCertificateRequest(cr).GetLeafCertificate(), nil
+		return NewCertificateRequestFromX509(cr).GetLeafCertificate(), nil
 	}
 
 	// With templates
--- a/x509util/certificate_request.go
+++ b/x509util/certificate_request.go
@@ -60,15 +60,15 @@ func NewCertificateRequest(signer crypto.Signer, opts ...Option) (*CertificateRe
 	return &cr, nil
 }
 
-// newCertificateRequest is an internal method that creates a CertificateRequest
-// from an x509.CertificateRequest.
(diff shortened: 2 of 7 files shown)


x509util/certificate.go
```diff
@@ -83,6 +83,83 @@ func NewCertificate(cr *x509.CertificateRequest, opts ...Option) (*Certificate,
...
+	c.PublicKeyAlgorithm = cert.PublicKeyAlgorithm
+
+	// Subject
+	c.Subject = newSubject(cert.Subject)
+
+	// SAN values
+	c.DNSNames = cert.DNSNames
+	c.EmailAddresses = cert.EmailAddresses
+	c.IPAddresses = cert.IPAddresses
+	c.URIs = cert.URIs
```
maraino, line 103: We should add the SANs too, as the device attestation certificates use another type of SANs by default. To do this, and probably to set the right DNSNames, EmailsAddresses, IPAddresses and URIs we will need to decode the subject alternative name extension (`2.5.29.17`).

maraino (changes requested):
We will need to add custom marshallers from a few types if they are not yet there.

Most of the certificate's important fields will already be in an `x509util.Certificate` format when it is constructed in `tls.go`. With templates, things like certificate modifiers and enforcers are used only sporadically, if they are used at all, to simplify things, we can just send that. I'm referring to `cert` in https://github.com/smallstep/certificates/blob/b2119e9f2c6852815971ba14585f1534240237e6/authority/tls.go#L137

A couple of things that won't be there are the NotAfter, NotBefore, and the provisioner extension.

maraino (commented):
LGTM if we remove `CertificateRequestFromX509` and renaming if necessary `newCertificateRequest`.

maraino:
@areed It looks like a change is now necessary in the workflows too.

maraino (approved):
lgtm

I didn't notice that we already have a NewCertificateRequest, it wasn't in the diff, either `NewCertificateRequestFromX509` or `CertificateRequestFromX509` would have been fine.

maraino (approved):
lgtm

PR #65, opened 2022-09-13
Update workflows to use common linter

dopey:


maraino (approved):
LGTM, but it looks like the linter is failing now

PR #68, opened 2022-09-23
[action] workflow improvements

dopey:
- Only run ci for master pushes and all pulls (removes duplicate workflows on push and pull)
- Add code scanning daily cron
- Add codeql and gitleaks workflows to CI run


.github/workflows/code-scan-cron.yml
```diff
@@ -0,0 +1,12 @@
...
+    - cron: '0 0 * * *'
+
+jobs:
+  gitleaks:
+    uses: smallstep/workflows/.github/workflows/gitleaks.yml@main
+    secrets:
+      GITLEAKS_LICENSE_KEY: ${{ secrets.GITLEAKS_LICENSE_KEY }}
+
+  codeql:
+    uses: smallstep/workflows/.github/workflows/codeql-analysis.yml@main
```
maraino, line 12: You just deleted that file? 

Also, is there any point in having all these fails in .github? Doesn't it make more sense to be directly on the root?

maraino: So the uses is simpler, `uses: smallstep/workflows/ci.yml`

maraino: Ah ok, this is crypto !!!!

maraino (approved):

PR #78, opened 2022-09-30
Marshal keys

maraino:
### Description

This PR adds MarshalPKIXPublicKey and MarshalPKCS8PrivateKey, these methods are used in `step` and should be available in `go.step.sm/crypto`.


maraino:
This is not necessary anymore.

PR #79, opened 2022-10-01
Fingerprints

maraino:
### Description

This PR extends the support for X.509 fingerprints to support all the options used by `step`. It also adds support for SSH fingerprints.

These functions will allow migrating more crypto-related code from `step` to `go.step.sm/crypto`


sshutil/fingerprint.go
```diff
@@ -0,0 +1,150 @@
...
+	"encoding/hex"
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	"go.step.sm/crypto/internal/emoji"
+	"golang.org/x/crypto/ssh"
+)
+
+// FingerprintEncoding defines the supported encodings in SSH key and
```
hslatman, line 17:
```suggestion
// FingerprintEncoding defines the supported encodings for SSH key and
```

sshutil/fingerprint.go
```diff
@@ -0,0 +1,150 @@
...
+
+// Fingerprint returns the SHA-256 fingerprint of an ssh public key or
+// certificate.
+func Fingerprint(pub ssh.PublicKey) string {
+	return EncodedFingerprint(pub, Base64RawFingerprint)
+}
+
+// EncodedFingerprint returns an encoded the SHA-256 fingerprint of an ssh
+// public key or certificate using the specified encoding. In an invalid
+// encoding is passed, the return value will be an empty string.
```
hslatman, line 48:
```suggestion
// EncodedFingerprint returns the SHA-256 hash of an ssh
// public key or certificate using the specified encoding. If an invalid
// encoding is passed, the return value will be an empty string.
```

sshutil/fingerprint.go
```diff
@@ -0,0 +1,150 @@
...
+		return prefix + emoji.Emoji(sum[:])
+	default:
+		return ""
+	}
+}
+
+// FormatFingerprint gets a public key from an authorized_keys file used in
+// OpenSSH and returns the fingerprint in the following format:
+//
+//	<size> SHA256:<base64-raw-fingerprint> <comment> (<type)
```
hslatman, line 74:
```suggestion
// FormatFingerprint returns an OpenSSH public key fingerprint in the following format:
//
//	<size> SHA256:<base64-raw-fingerprint> <comment> (<type)
```

sshutil/fingerprint.go
```diff
@@ -0,0 +1,150 @@
...
+		size = 8 * k.Size()
+	case ssh.KeyAlgoDSA:
+		typ = "DSA"
+		cpk, err := CryptoPublicKey(key)
+		if err != nil {
+			return "", 0, err
+		}
+		k, ok := cpk.(*dsa.PublicKey)
+		if !ok {
+			return "", 0, errors.New("unsupported key: not an DSA public key")
```
hslatman, line 138:
```suggestion
			return "", 0, errors.New("unsupported key: not a DSA public key")
```

sshutil/fingerprint.go
maraino: Fixed with 613cf63

sshutil/fingerprint.go
maraino: Fixed with 613cf63, but I've haven't used your suggestion.

sshutil/fingerprint.go
maraino: Fixed with 613cf63

hslatman (approved):

sshutil/fingerprint.go
hslatman: Yeah, I'm OK with the new one. At first it read like the code would also read `authorized_keys`, which it obviously didn't. Now it makes sense, as it's more of a description of what type of public key it is.

hslatman:
I think most of the code came from `smallstep/cli` already, right?

maraino:
> I think most of the code came from smallstep/cli already, right? 

Yes

PR #86, opened 2022-10-04
Add Encrypt helper

maraino:
### Description

This PR adds the new method `Encrypt`, which encrypts data using `PBES2-HS256+A128KW`. This method is a generalization of the already existing `EncryptJWK`, but it gets the raw data and `joseOption` as input, in the same fashion as the `Decrypt` method.

This new method allows easier integration with `step`, with an option is easier to pass a password prompter or a password file.


jose/encrypt.go
```diff
@@ -19,12 +19,30 @@ type PasswordPrompter func(s string) ([]byte, error)
...
 	if err != nil {
-		return nil, errors.Wrap(err, "error marshaling JWK")
+		return nil, err
+	}
+
+	var passphrase []byte
+	switch {
+	case len(ctx.password) > 0:
+		passphrase = ctx.password
+	case ctx.passwordPrompter != nil || PromptPassword != nil:
```
areed, line 34: This would be simpler with two separate cases

jose/encrypt.go
```diff
@@ -40,22 +58,35 @@ func EncryptJWK(jwk *JSONWebKey, passphrase []byte) (*JSONWebEncryption, error)
...
 
-	opts := new(EncrypterOptions)
-	opts.WithContentType(ContentType("jwk+json"))
+	encrypterOptions := new(EncrypterOptions)
+	if ctx.contentType != "" {
+		encrypterOptions.WithContentType(ContentType(ctx.contentType))
+	}
 
-	encrypter, err := NewEncrypter(DefaultEncAlgorithm, recipient, opts)
+	encrypter, err := NewEncrypter(DefaultEncAlgorithm, recipient, encrypterOptions)
```
areed, line 66: Is there any benefit of using AES with a 256 bit key here when that key is encrypted with a 128 bit key?

areed (dismissed):

jose/encrypt.go
maraino: I was fixing a linter error, which was the direct fix. 
Fixed with c855ec6

jose/encrypt.go
maraino: Probably not, but that has been the default behavior for a long time, take a look at any JWK provisioner:

```shell
$ echo <encrypted-key> | step crypto jose format | jq -r .protected | step base64 -d | jq
{
  "alg": "PBES2-HS256+A128KW",
  "cty": "jwk+json",
  "enc": "A256GCM",
  "p2c": 100000,
  "p2s": "<16-byte-salt>"
}
```

The same applies to the default encryption algorithm when you use `step crypto jwe encrypt` with any password-base algorithm.

If we want to make any change would be to `PBES2-HS512+A256KW` by default. In any case, the key with more risk is actually the password.

areed (dismissed):

areed (approved):

PR #92, opened 2022-10-06
YubiKey pin and touch policies

maraino:
### Description

The YubiKey KMS allows to set pin and touch policies when a key is created, these policies are used when signing or decrypting with an asymmetric key in a slot.

The default pin policy is set to always and the touch policy to never, but now the values can be configurable.

`step-kms-plugin` will add `--pin-policy` and `--touch-policy` flags to be able to configure these.

### DIFF
--- a/kms/apiv1/requests.go
+++ b/kms/apiv1/requests.go
@@ -19,6 +19,35 @@ const (
 	HSM
 )
 
+// PINPolicy represents PIN requirements when signing or decrypting with an
+// asymmetric key in a given slot. PINPolicy is used by the YubiKey KMS.
+type PINPolicy int
+
+// PIN policies supported by this package. The values must match the ones in
+// github.com/go-piv/piv-go/piv.
+//
+// Caching for PINPolicyOnce isn't supported on YubiKey
+// versions older than 4.3.0 due to issues with verifying if a PIN is needed.
+// If specified, a PIN will be required for every operation.
+const (
+	PINPolicyNever PINPolicy = iota + 1
+	PINPolicyOnce
+	PINPolicyAlways
+)
+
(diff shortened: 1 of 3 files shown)


kms/apiv1/requests.go
```diff
@@ -19,6 +19,35 @@ const (
 	HSM
 )
 
+// PINPolicy represents PIN requirements when signing or decrypting with an
+// asymmetric key in a given slot. PINPolicy is used by the YubiKey KMS.
+type PINPolicy int
+
+// PIN policies supported by this package. The values must match the ones in
+// github.com/go-piv/piv-go/piv.
```
hslatman, line 27: If they have to match, they could be aliased to the types in `go-piv`?

kms/apiv1/requests.go
```diff
@@ -19,6 +19,35 @@ const (
...
+// versions older than 4.3.0 due to issues with verifying if a PIN is needed.
+// If specified, a PIN will be required for every operation.
+const (
+	PINPolicyNever PINPolicy = iota + 1
+	PINPolicyOnce
+	PINPolicyAlways
+)
+
+// TouchPolicy represents proof-of-presence requirements when signing or
+// decrypting with asymmetric key in a given slot. PINPolicy is used by the
```
hslatman, line 39:
```suggestion
// decrypting with asymmetric key in a given slot. TouchPolicy is used by the
```

areed (dismissed):

kms/apiv1/requests.go
maraino: @hslatman I don't want to depend on go-piv and cgo in this package.

hslatman (approved):

PR #101, opened 2022-10-18
[kms/yubikey] Add decryption support to Yubikey KMS plugin

fuhry:
Some provisioners in `step-ca` (i.e. SCEP) require the ability to decrypt challenges using the intermediate CA key. [1] Although the Yubikey hardware supports decryption using RSA keys, support for this was missing from the Yubikey KMS plugin.

This commit implements the `CreateDecrypter` method in the Yubikey KMS plugin. Tests are included.

[1] https://github.com/smallstep/certificates/blob/361e2b2907bc94744da98ece290fdc1d851951f4/authority/authority.go#L540-L545

Signed-Off-By: Dan Fuhry <dan@fuhry.com>

### DIFF
--- a/kms/yubikey/yubikey.go
+++ b/kms/yubikey/yubikey.go
@@ -219,6 +219,42 @@ func (k *YubiKey) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, e
 	return signer, nil
 }
 
+// CreateDecrypter creates a crypto.Decrypter using the key present in the configured
+// Yubikey slot.
+func (k *YubiKey) CreateDecrypter(req *apiv1.CreateDecrypterRequest) (crypto.Decrypter, error) {
+	slot, err := getSlot(req.DecryptionKey)
+	if err != nil {
+		return nil, err
+	}
+
+	pin := k.pin
+	if pin == "" {
+		// Attempt to get the pin from the uri
+		if u, err := uri.ParseWithScheme(Scheme, req.DecryptionKey); err == nil {
+			pin = u.Pin()
+		}
+	}
+
+	pub, err := k.getPublicKey(slot)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


fuhry:
TODO: the CreateSigner and CreateDecrypter methods could be DRYed out a bit

maraino:
Nice, I wanted to do this at some point.

kms/yubikey/yubikey.go
```diff
@@ -219,6 +219,42 @@ func (k *YubiKey) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, e
...
+		return nil, err
+	}
+
+	priv, err := k.yk.PrivateKey(slot, pub, piv.KeyAuth{
+		PIN:       pin,
+		PINPolicy: piv.PINPolicyAlways,
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "error retrieving private key")
+	}
```
maraino, line 249: Although `CreateSigner` is implemented this way, in other KMSs, to create a signer, you need to `CreateKey -> CreateSigner`. We should use the same logic for `CreateDecrypter`, `CreateKey -> CreateDecrypter`.

I'll add a PR to fix `CreateSigner`

maraino (changes requested):

kms/yubikey/yubikey.go
maraino: No, I though this method was generating the key, I'm wrong.

fuhry:
Thanks for merging this!

It turns out that this PR on its own is not enough. The `*pki.keyRSA` (cast to `crypto.Decrypter`) returned from this doesn't match the cases checked in [pkcs7/decrypt.go](https://github.com/smallstep/pkcs7/blob/5660c44b3d3b8a1c4a125dee9ecfd658efb3f437/decrypt.go#L34), so [here's a PR to fix that](https://github.com/smallstep/pkcs7/pull/3).

PR #103, opened 2022-10-19
WIP: Initial CAPI KMS implementation

unreality:
#### Name of feature:

Adds a KMS that uses Windows CAPI/nCrypt as its key store

First step for https://github.com/smallstep/cli/issues/767

Windows 10+ only, and not in WSL

Non-Windows platforms

💔Thank you!


unreality:
This is pretty rough and has no tests at the moment, but the functionality is all there.

Some things that are not working:
* When using the kms to store certificates, it flows through to `LoadCertificate` to output the certificate after it is stored. Since the SHA1 fingerprint is not given during `StoreCertificate`, printing the certificate fails even though the certificate import succeeded. You can pass in the sha1 when running the import, but it seems like a bad experience for the user to have to specify the SHA1 thumbprint when importing a certificate to avoid an error.
* If you create a key on a smart card and then import a certificate, the plugin fails to associate the private key correctly.

I'm unsure whether parsing the URI for `CreateSigner`, `CreateKey`, etc is the preferred way of doing things, it seems that what the other KMS libs do...?

kms/capi/capi.go
```diff
@@ -0,0 +1,565 @@
...
+		// generate a uuid for the container name
+		containerUUID, err := uuid.NewRandom()
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to generate uuid")
+		}
+		k.containerName = containerUUID.String()
+	}
+
+	alg, ok := signatureAlgorithmMapping[req.SignatureAlgorithm]
+
```
maraino, line 254:
```suggestion
```

kms/capi/capi.go
```diff
@@ -0,0 +1,565 @@
...
+	// create a CRYPT_INTEGER_BLOB -- https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/aa381414(v=vs.85)
+	hashBlob := struct {
+		len  uint32
+		hash uintptr
+	}{
+		len:  uint32(len(sha1Bytes)),
+		hash: uintptr(unsafe.Pointer(&sha1Bytes[0])),
+	}
+
+	certHandle, err := FindCertificateInStore(st,
```
maraino, line 397: In other methods, you are freeing some handles, do we need to free this one?

kms/capi/capi.go
```diff
@@ -0,0 +1,565 @@
...
+
+type CAPISigner struct {
+	algorithmGroup string
+	providerHandle uintptr
+	keyHandle      uintptr
+	containerName  string
+	PublicKey      crypto.PublicKey
+}
+
+func NewCAPISigner(providerHandle uintptr, containerName string, pin string) (crypto.Signer, error) {
```
maraino, line 488: This should be private.

kms/capi/ncrypt_windows.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+// wide returns a pointer to a a uint16 representing the equivalent
+// to a Windows LPCWSTR.
+func wide(s string) *uint16 {
+	w, _ := syscall.UTF16PtrFromString(s)
+	return w
+}
+
+func NCryptOpenStorage(provider string) (uintptr, error) {
```
maraino, line 145: All or most of these methods should be private for now unless you are using this package differently.

maraino (changes requested):
Hi @unreality, this is awesome, I've added a "few" comments that should be easy to fix

kms/capi/capi.go
```diff
@@ -0,0 +1,565 @@
...
+	"unsafe"
+)
+
+// Scheme is the scheme used in uris.
+const Scheme = "capi"
+
+const (
+	ProviderNameArg  = "provider"
+	ContainerNameArg = "key"
+	HashArg          = "sha1"
```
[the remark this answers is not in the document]
unreality: This is the hash to find the certificate within the Windows certificate store, unfortunately it only uses sha1 as its fingerprint algorithm. We can search by other things like serial number etc, but for now i've just implemented sha1 searching.

kms/capi/capi.go
```diff
@@ -0,0 +1,565 @@
...
+		certStoreLocation = certStoreLocalMachine
+	default:
+		return nil, errors.Errorf("invalid cert store location %v", storeLocation)
+	}
+
+	var storeName string
+
+	// default to storing in the 'My' store
+	if storeName = u.Get(StoreNameArg); storeName == "" {
+		storeName = "My"
```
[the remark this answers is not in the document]
unreality: Yes, 'My' is the name of the personal certificate store. There are a few other defaults like 'Root' and 'CA' and users are able to create their own if they wish. The personal store is generally where people want to put certs and keys though.

kms/capi/capi.go
```diff
@@ -0,0 +1,565 @@
...
+		if !ok {
+			return nil, fmt.Errorf("unsupported RSA hash algorithm %v", hf)
+		}
+		signatureBytes, err := NCryptSignHash(s.keyHandle, digest, hashAlg)
+
+		if err != nil {
+			return nil, errors.Wrap(err, "NCryptSignHash failed")
+		}
+
+		return signatureBytes, nil
```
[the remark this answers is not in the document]
unreality: The list of supported key algorithms doesnt mention RSA-PSS (https://learn.microsoft.com/en-us/windows/win32/seccng/cng-algorithm-identifiers) but that site does mention third-party plugins, and there is some info on using a TPM 2.0 to make RSA-PSS hashes. I'll look into it more but i think at the moment it doesnt look like its supported in the standard CAPI lib. 

Of course its possible i've misunderstood RSA-PSS and maybe it is supported, i'll have to do some more reading about it :)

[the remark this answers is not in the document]
maraino: It looks like only PKCS#1 is supported by default. 

| Constant/value | Description |
| -- | -- |
| BCRYPT_RSA_ALGORITHM"RSA" | The RSA public key algorithm. Standard: PKCS #1 v1.5 and v2.0. |
| BCRYPT_RSA_SIGN_ALGORITHM"RSA_SIGN" | The RSA signature algorithm. This algorithm is not currently supported. You can use the BCRYPT_RSA_ALGORITHM algorithm to perform RSA signing operations. Standard: PKCS #1 v1.5 and v2.0. |

We should fail if `_, ok := opts(*rsa.PSSOptions)`, at least for now.

By the way, adding the decrypter interface would be nice, is optional but can be used with SCEP, and with this https://github.com/smallstep/step-kms-plugin/pull/14:
```go
// Decrypter is an interface implemented by KMSes that are used
// in operations that require decryption
type Decrypter interface {
	CreateDecrypter(req *CreateDecrypterRequest) (crypto.Decrypter, error)
}
```

And it looks like it supports both RSA-OAEP and RSA PKCS#1 v1.5. It should be part of a different PR if you're up to:
https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptdecrypt

kms/capi/capi.go
```diff
@@ -0,0 +1,646 @@
...
+	if k.providerHandle != 0 {
+		return nCryptFreeObject(k.providerHandle)
+	}
+
+	return nil
+}
+
+// CreateSigner returns a new signer configured with the given signing key.
+func (k *CAPIKMS) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, error) {
+	var u *uri.URI
```
maraino, line 275:
```suggestion
	if req.SigningKey == "" {
		return nil, errors.New("signing key cannot be empty")
	}
```

kms/capi/capi.go
```diff
@@ -0,0 +1,646 @@
...
+		scPIN = k.pin
+	}
+
+	return newCAPISigner(k.providerHandle, k.containerName, scPIN)
+}
+
+// CreateKey generates a new key using Golang crypto and returns both public and
+// private key.
+func (k *CAPIKMS) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyResponse, error) {
+
```
maraino, line 298:
```suggestion
```

kms/capi/capi.go
```diff
@@ -0,0 +1,646 @@
...
+	if req.Name == "" {
+		return nil, errors.Errorf("capi signing request must have a name")
```
maraino, line 300:
```suggestion
		return nil, errors.New("createKeyRequest 'name' cannot be empty")
```

kms/capi/capi.go
```diff
@@ -0,0 +1,646 @@
...
+		return nil, fmt.Errorf("failed to parse URI: %w", err)
+	}
+
+	if k.containerName = u.Get(ContainerNameArg); k.containerName == "" {
+		// generate a uuid for the container name
+		k.containerName, err = randutil.UUIDv4()
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate uuid: %w", err)
+		}
+	}
```
maraino, line 314: This is weird, you should not update the `containerName` initialized in `New()`. I'm not sure but it looks like `k.containerName` might not be necessary and you just need to create one if the uri doesn't provide it.

If you need the same container name in `GetPublicKey` or `CreateSigner`, then the return `Name` should be an uri with all the necessary information to use it.

In general, the return should be something like this:

```
return &apiv1.CreateKeyResponse{
 	Name:      "capi:key=foo;foo=bar;...",
 	PublicKey: pub,
 	CreateSignerRequest: apiv1.CreateSignerRequest{
 		SigningKey: "capi:key=foo;foo=bar;...",
 	},
 }, nil
 ```

File: kms/capi/capi.go [hunk 12]
```diff
@@ -0,0 +1,646 @@
...
+		return nil, fmt.Errorf("unable to retrieve public key: %w", err)
+	}
+
+	return &apiv1.CreateKeyResponse{
+		Name:      uc,
+		PublicKey: pub,
+		CreateSignerRequest: apiv1.CreateSignerRequest{
+			SigningKey: uc,
+		},
+	}, nil
```
maraino, line 387: See my previous comment.

kms/capi/capi.go
```diff
@@ -0,0 +1,646 @@
...
+	scPIN := u.Pin()
+
+	//failover to pin set in kms instantiation
+	if scPIN == "" {
+		scPIN = k.pin
+	}
+
+	if scPIN != "" {
+		err = nCryptSetProperty(kh, NCRYPT_PIN_PROPERTY, scPIN, 0)
+
```
maraino, line 360:
```suggestion
```

kms/capi/capi.go
```diff
@@ -0,0 +1,646 @@
...
+		CreateSignerRequest: apiv1.CreateSignerRequest{
+			SigningKey: uc,
+		},
+	}, nil
+}
+
+// GetPublicKey returns the public key from the file passed in the request name.
+func (k *CAPIKMS) GetPublicKey(req *apiv1.GetPublicKeyRequest) (crypto.PublicKey, error) {
+	var u *uri.URI
+
```
maraino, line 393:
```suggestion
```

kms/capi/capi.go
```diff
@@ -0,0 +1,646 @@
...
+	u, err := uri.ParseWithScheme(Scheme, req.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse URI: %w", err)
+	}
+
+	if k.containerName = u.Get(ContainerNameArg); k.containerName == "" {
+		return nil, fmt.Errorf("GetPublicKey %v not specified", ContainerNameArg)
+	}
```
maraino, line 401: Do not overwrite containerName, see my previous comment. This should not be a field of CAPIKMS, it should always be in the URI.

kms/capi/capi.go
```diff
@@ -0,0 +1,646 @@
...
+func (k *CAPIKMS) LoadCertificate(req *apiv1.LoadCertificateRequest) (*x509.Certificate, error) {
+	u, err := uri.ParseWithScheme(Scheme, req.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse URI: %w", err)
+	}
+
+	var sha1Hash string
+	if sha1Hash = u.Get(HashArg); sha1Hash == "" {
+		return nil, fmt.Errorf("%v is required", HashArg)
+	}
```
maraino, line 422: You mentioned before that you could also look by serial number, right? 

If this is possible, I would probably add both options. If `sha1` is present, use it, if not `serial`, and fail if both are not present. 

If both are present and you retrieve it by sha1, you should fail if the serial does not match.

maraino (changes requested):
I added a few other comments, you should use URIs in CreateKey response.

kms/capi/capi.go
unreality: ive added searching for issuer+serial, and subject key id

unreality:
Just adding a note that nCrypt looks like it can create attestation claims, so the KMS might be able to support the attest interface at some point (https://learn.microsoft.com/en-us/windows/win32/api/ncrypt/nf-ncrypt-ncryptcreateclaim)

kms/capi/capi.go
```diff
@@ -0,0 +1,715 @@
...
+		}
+
+		// iterate over all certificates from issuer, and check the SN
+		var prevCert *windows.CertContext = nil
+		for {
+			certHandle, err = findCertificateInStore(st,
+				encodingX509ASN|encodingPKCS7,
+				0,
+				findIssuerStr,
+				uintptr(unsafe.Pointer(wide(issuerName))), prevCert)
```
maraino, line 541: Can't we use [CERT_ISSUER_SERIAL_NUMBER](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_issuer_serial_number) instead? It is one of the members of the [CERT_ID](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_id) union that we use above. It will make the for loop unnecessary, as according to the docs "The combination of the issuer name and the serial number is a unique identifier of a certificate."

Do you know what is exactly the "issuer name"? The full issuer `C=US,ST=California,L=San Francisco,O=GitHub, Inc.,CN=github.com`? The CommonName?

kms/capi/capi.go
```diff
@@ -0,0 +1,715 @@
...
+	SerialNumberArg  = "serial"
+	IssuerNameArg    = "issuer"
+)
+
+var signatureAlgorithmMapping = map[apiv1.SignatureAlgorithm]string{
+	apiv1.UnspecifiedSignAlgorithm: "ECDSA_P256",
+	apiv1.SHA256WithRSA:            "RSA",
+	apiv1.SHA512WithRSA:            "RSA",
+	apiv1.SHA256WithRSAPSS:         "RSA",
+	apiv1.SHA512WithRSAPSS:         "RSA",
```
maraino, line 49: We should remove this, as we're currently not supporting them. 

And you should add `apiv1.SHA384WithRSA`, as we are supporting them. Although for this specific KMS it doesn't matter which we want to use, we should be consistent with the API.

kms/capi/ncrypt_windows.go
```diff
@@ -0,0 +1,505 @@
...
+		0, // pvReserved
+		0, // pbEncoded
+		uintptr(unsafe.Pointer(&size)),
+		0,
+	)
+	if r != 1 {
+		return nil, fmt.Errorf("CertStrToName returned %v during size check: %w", errNoToStr(uint32(r)), err)
+	}
+
+	fmt.Printf("certStrToName size was %d\n", size)
```
maraino, line 489: Remove debug printf.

maraino (changes requested):
Hi @unreality, I've added a few more comments.

Can you re-request my review when you make the changes, I usually don't look at them unless I see it as a pending review.

kms/capi/capi.go
unreality: I would love to be able to use [CERT_ISSUER_SERIAL_NUMBER](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_issuer_serial_number), but i spent quite a few hours not being able to. I believe there is something weird where that search criteria needs absolutely exact matches, but CERT_FIND_ISSUER_STR does something else which allows fuzzier matches?

So to make it work i used the combination for now, its sub-optimal but I couldnt find any resources or examples on it to guide me and I wasnt making any progress with it unfortunately.

the "issuer name" is the full `pkix.Name`

kms/capi/capi.go
```diff
@@ -0,0 +1,715 @@
...
+		}
+
+		if certHandle == nil {
+			return nil, fmt.Errorf("certificate with %v=%s not found", KeyIDArg, keyID)
+		}
+
+		defer windows.CertFreeCertificateContext(certHandle)
+		return certContextToX509(certHandle)
+	} else if issuerName != "" && serialNumber != "" {
+		serialBytes, err := hex.DecodeString(serialNumber)
```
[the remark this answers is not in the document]
unreality: Given windows reports all its serial numbers in hex, perhaps we could look at the string and if it starts with 0x, allow hex values?

Its not a great user experience for windows users if they need to convert it to a different format i think...

kms/capi/capi.go
maraino: I haven't tested this, but this Russian forum might provide some insight
https://www.cryptopro.ru/forum2/default.aspx?g=posts&t=15614

kms/capi/capi.go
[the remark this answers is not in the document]
maraino: Sure, hex encoding starts with `0x`, decimal encoding otherwise.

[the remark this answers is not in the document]
maraino: step certificate inspect shows both:
```console
$ step certificate inspect mariano.crt | grep "Serial Number"
        Serial Number: 130456639100600112210690927965585162120 (0x622508e69b9ba98490f9c1a5601e3f88)
```

maraino (commented):
Hi @unreality, have you tried to solve the [CERT_ISSUER_SERIAL_NUMBER](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_issuer_serial_number) thing?

And can you install [golangci-lint](https://github.com/golangci/golangci-lint) and run `make lint` on your windows environment?

unreality:
Im still working on the [CERT_ISSUER_SERIAL_NUMBER](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_issuer_serial_number) problem. There are a few other minor things i'd like to try fix as well, so i'll request a review once i've got all that sorted.

I actually have golangci-lint installed, but the set of checks thats configured for this project needs some extra work to enable :) I'll put the effort in to get the GNU diff tool installed

unreality:
I'm cleaning up this a bit and have come across a puzzling issue;

If you use provider=`Microsoft Platform Crypto Provider`, the KMS plugin cannot open the provider. The two other providers `Microsoft Software Key Storage Provider` and `Microsoft Smart Card Key Storage Provider` both work as expected, and if you create and run:
```
func main() {
	p, e := nCryptOpenStorage("Microsoft Platform Crypto Provider")
	if e != nil { panic(p) }
        fmt.Printf("Success\n")
	nCryptFreeObject(p)
}
```

Everything works fine.

@maraino are you aware of any thread-based things that occur in `cobra` or the KMSFS that might be interfering with getting a pointer correctly? I know its a bit of a long shot, but im unsure whats interfering with `nCryptOpenStorage`

unreality:
I've fixed up a few things and tidied up some stuff, but i still cannot work out [CERT_ISSUER_SERIAL_NUMBER](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_issuer_serial_number), i can successfully search for Issuer Name using an X.500 string, but as soon as I add the serial number it doesnt find anything. So ive just left it as the inefficient search for now

kms/capi/ncrypt_windows.go
```diff
@@ -0,0 +1,572 @@
...
+import (
+	"bytes"
+	"crypto"
+	"crypto/elliptic"
+	"crypto/sha1"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"golang.org/x/sys/windows"
+	"syscall"
```
maraino, line 15: I believe all calls to syscall can be just converted to windows, the original syscall is deprecated.

```go
windows.UTF16PtrFromString(s)
windows.Errno(0)
```

kms/capi/capi.go
```diff
@@ -0,0 +1,754 @@
...
+)
+
+var signatureAlgorithmMapping = map[apiv1.SignatureAlgorithm]string{
+	apiv1.UnspecifiedSignAlgorithm: "ECDSA_P256",
+	apiv1.SHA256WithRSA:            "RSA",
+	apiv1.SHA384WithRSA:            "RSA",
+	apiv1.SHA512WithRSA:            "RSA",
+	apiv1.ECDSAWithSHA256:          "ECDSA_P256",
+	apiv1.ECDSAWithSHA384:          "ECDSA_P384",
+	apiv1.ECDSAWithSHA512:          "ECDSA_P521",
```
maraino, line 52: You have constants in ncrypt_windows.go for those strings. It would make sense to use them here too.

kms/capi/capi.go
```diff
@@ -0,0 +1,754 @@
...
+		pwHash, err := hashPasswordUTF16(pinOrPass) // we have to SHA1 hash over the utf16 string
+
+		if err != nil {
+			return nil, fmt.Errorf("unable to hash pin: %w", err)
+		}
+		err = nCryptSetProperty(kh, NCRYPT_PCP_USAGE_AUTH_PROPERTY, pwHash, 0)
+		if err != nil {
+			return nil, fmt.Errorf("unable to set key NCRYPT_PIN_PROPERTY: %w", err)
+		}
+	}
```
maraino, line 421: Is there a missing else here?

kms/capi/capi.go
```diff
@@ -0,0 +1,754 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to hash password: %w", err)
+		}
+
+		err = nCryptSetProperty(kh, NCRYPT_PCP_USAGE_AUTH_PROPERTY, passHash, 0)
+
+		if err != nil {
+			return nil, fmt.Errorf("unable to set key NCRYPT_PCP_USAGE_AUTH_PROPERTY: %w", err)
+		}
+	}
```
maraino, line 331: It looks like that an `else` is missing here.

kms/capi/capi.go
```diff
@@ -0,0 +1,754 @@
...
+			uintptr(unsafe.Pointer(&searchData)), nil)
+		if err != nil {
+			return nil, fmt.Errorf("findCertificateInStore failed: %w", err)
+		}
+		if certHandle == nil {
+			return nil, fmt.Errorf("certificate with %v=%s not found", KeyIDArg, keyID)
+		}
+		defer windows.CertFreeCertificateContext(certHandle)
+		return certContextToX509(certHandle)
+	case issuerName != "" && serialNumber != "":
```
maraino, line 574: Ok, but add a TODO about [CERT_ISSUER_SERIAL_NUMBER](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_issuer_serial_number)

kms/capi/capi.go
```diff
@@ -0,0 +1,754 @@
...
+			bi := new(big.Int)
+			bi, ok := bi.SetString(serialNumber, 10)
+			if !ok {
+				return nil, fmt.Errorf("invalid %v - must be in hex or integer format", SerialNumberArg)
+			}
+			serialBytes = bi.Bytes()
+		}
+		var prevCert *windows.CertContext
+		for {
+			certHandle, err = findCertificateInStore(st,
```
maraino, line 593: In the previous cases, you are freeing the handle with `windows.CertFreeCertificateContext(certHandle)`, shouldn't we free them here too? Note that you're iterating using them.

maraino (changes requested):
Pretty much ok, but there are a couple of missing `else` and some handles that are not free.

You should also always use `x/sys/windows` instead of `syscall`.

Can you also merge/rebase master to fix the conflict.

kms/capi/capi.go
unreality: if not supplied a UI prompt will be shown to the user, and at the moment we can only set the pin/pass for MSSC and MSPCP providers

kms/capi/capi.go
unreality: @maraino depends if you want the key creation to fail when the user provides a pin but specifies the MSKSP (ie software) provider?

kms/capi/capi.go
unreality: The handle is freed below on L615

As far as im aware `findCertificateInStore` frees `prevCert` before returning a new one, so we only need to free the handle if `findCertificateInStore` returns us something and we DONT continue iterating through (ie we find our cert).

If `findCertificateInStore` exhausts its search it frees the handle and returns nil

kms/capi/capi.go
maraino: Well, I trust your judgment here. We can allow the ProviderMSKSP with a PIN even if it never requires one.

kms/capi/capi.go
maraino: This is ok in most cases.

kms/capi/capi.go
unreality: @maraino I can change the search from using `findIssuerStr` to `findIssuerName` which will require users to be explicit about the issuer name and supply it in X.500 format if thats more acceptable?

That SHOULD mean any future fix will only need to solve the SerialNumber search -- i believe its looking for a format that im not aware of

maraino:
@unreality Ready for review?

maraino (approved):
Thanks again @unreality.

I'll create a new tag and create a new release of step-kms-plugin when Go 1.19.3 is available in the CI.

PR #105, opened 2022-10-24
WIP: Allow x5c to use KMS certs and keys for token generation

unreality:
Required changes for https://github.com/smallstep/cli/pull/783

Im unsure how to check if a public key matches a private one when we dont have access to the private key data, so that verification is skipped at the moment.

💔Thank you!

### DIFF
--- a/jose/validate.go
+++ b/jose/validate.go
@@ -8,6 +8,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"fmt"
+	"gopkg.in/square/go-jose.v2"
 	"os"
 
 	"github.com/pkg/errors"
@@ -49,8 +50,12 @@ func validateX5(certs []*x509.Certificate, key interface{}) error {
 		return errors.New("certs cannot be empty")
 	}
 
-	if err := keyutil.VerifyPair(certs[0].PublicKey, key); err != nil {
-		return errors.Wrap(err, "error verifying certificate and key")
+	// We cant get the private key data to directly compare, so don't actually check the keys match if we have
+	// an OpaqueSigner. Perhaps the public key can be derived from the OpaqueSigner and matched?
+	if _, isOpaqueSigner := key.(jose.OpaqueSigner); !isOpaqueSigner {
+		if err := keyutil.VerifyPair(certs[0].PublicKey, key); err != nil {
+			return errors.Wrap(err, "error verifying certificate and key")
+		}
 	}
 
 	if certs[0].KeyUsage&x509.KeyUsageDigitalSignature == 0 {
--- a/jose/types.go
+++ b/jose/types.go
@@ -3,7 +3,9 @@
 package jose
 
 import (
+	"crypto"
 	"errors"
+	"gopkg.in/square/go-jose.v2/cryptosigner"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


jose/validate.go
```diff
@@ -49,8 +50,12 @@ func validateX5(certs []*x509.Certificate, key interface{}) error {
 		return errors.New("certs cannot be empty")
 	}
 
-	if err := keyutil.VerifyPair(certs[0].PublicKey, key); err != nil {
-		return errors.Wrap(err, "error verifying certificate and key")
+	// We cant get the private key data to directly compare, so don't actually check the keys match if we have
+	// an OpaqueSigner. Perhaps the public key can be derived from the OpaqueSigner and matched?
+	if _, isOpaqueSigner := key.(jose.OpaqueSigner); !isOpaqueSigner {
```
maraino, line 55: We can use our alias:

```suggestion
	if _, isOpaqueSigner := key.(OpaqueSigner); !isOpaqueSigner {
```

maraino: We should match the public key. `*rsa.PublicKey`, `*ecdsa.PublicKey` and `ed25519.PublicKey` implements `Equal(x crypto.PublicKey) bool`

maraino (changes requested):
Looks good, but we should make sure the keys match.

jose/validate.go
```diff
@@ -49,8 +50,33 @@ func validateX5(certs []*x509.Certificate, key interface{}) error {
...
+		switch pub := certs[0].PublicKey.(type) {
+		case *rsa.PublicKey:
+			publicKeysMatch = pub.Equal(signerPub)
+		case *ecdsa.PublicKey:
+			publicKeysMatch = pub.Equal(signerPub)
+		case ed25519.PublicKey:
+			publicKeysMatch = pub.Equal(signerPub)
+		default:
+			return errors.Errorf("unsupported public key type %T", pub)
+		}
```
maraino, line 71: PublicKey is an empty interface, but you should be able to create a private interface and test against it. 

You can even an anonymous one:

```go
pub, ok := certs[0].PublicKey.((interface{ Equal(crypto.PublicKey) bool })
if ok {
	publicKeysMatch = pub.Equal(signerPub)
} else {
	return errors.Errorf("unsupported public key type %T", certs[0].PublicKey)
}
```

maraino (commented):
Can you use a typed or anonymous interface to avoid that "annoying" thing?

A test would be nice too.

jose/validate.go
maraino: @unreality take a look at the linter errors too.

unreality: > PublicKey is an empty interface, but you should be able to create a private interface and test against it.

Neat, thanks for the tip :)

maraino: @unreality do you want to do this change?

unreality: @maraino i thought it was done in commit [6afc19d](https://github.com/smallstep/crypto/pull/105/commits/6afc19dfc9c4b38250c1572d00beafef94550270) ?

Have i messed up? :)

maraino: I messed up, I clicked on the [jose/validate.go](https://github.com/smallstep/crypto/pull/105/files/ceedd033352f40cfff1455eb1f28d5fd92610990#diff-019ea4b9c6a1e4a6a5d2054d144cc4c9db037800ce6f3aae0efe90480596d862) of this comment and showed me the old version.

maraino (approved):
👍

PR #115, opened 2022-11-02
Use Equal to verify a keypair

maraino:
### Description

This PR changes the keypair verification to use the Equal method. Now the RSA key exponent is also verified, that was a problem with the previous version.

This method can also be simplified using crypto.Signer and an "Equaler" interface, but I've decided to leave it as it currently is.

### DIFF
--- a/jose/types.go
+++ b/jose/types.go
@@ -5,12 +5,12 @@ package jose
 import (
 	"crypto"
 	"errors"
-	"gopkg.in/square/go-jose.v2/cryptosigner"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


maraino:
@unreality I've fixed your `NewOpaqueSigner`, see https://github.com/smallstep/crypto/pull/115/commits/4c1fc2efcf0339bf92f19dc9183edfdb00bb8561

I'm also considering using crypto.Signer on keyutil.VerifyKeypair, so the if condition in validateX5 is not necessary.

dopey (approved):

dopey (approved):

PR #166, opened 2023-02-01
Add tags to remove KMSs at compile time

### LINKED ISSUE #163: Build tags to disable kms types
### What would you like to be added

Allow using Go's build tags to disable a KMS.

### Why this is needed

This allows, for example, to not require `pcsc-lite` if we want to compile `step-ca` with just PKCS#11 support.

For example, if we want to disable YubiKey support only we could use something like:

```console
go build -tags noyubikey -o bin/step-ca cmd/step-ca/main.go
```

maraino:
### Description

This PR adds support for passing the following tags to exclude a specific KMS from a binary:
 - noawskms
 - noazurekms
 - nocapi
 - nocloudkms
 - nopkcs11
 - nosshagentkms
 - noyubikey

Fixes #163

### DIFF
--- a/kms/capi/no_capi.go
+++ b/kms/capi/no_capi.go
@@ -1,0 +1,20 @@
+//go:build !windows || nocapi
+// +build !windows nocapi
+
+package cloudkms
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"go.step.sm/crypto/kms/apiv1"
+)
+
+func init() {
+	apiv1.Register(apiv1.CAPIKMS, func(ctx context.Context, opts apiv1.Options) (apiv1.KeyManager, error) {
+		name := filepath.Base(os.Args[0])
+		return nil, errors.Errorf("unsupported kms type 'capi': %s is compiled without Microsoft CryptoAPI support", name)
+	})
+}
--- a/kms/yubikey/yubikey_no_cgo.go
+++ b/kms/yubikey/yubikey_no_cgo.go
@@ -1,5 +1,5 @@
-//go:build !cgo
-// +build !cgo
+//go:build !cgo || noyubikey
+// +build !cgo noyubikey
 
 package yubikey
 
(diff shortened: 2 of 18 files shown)


kms/yubikey/yubikey_no_cgo.go
```diff
@@ -15,6 +15,6 @@ import (
 func init() {
 	apiv1.Register(apiv1.YubiKey, func(ctx context.Context, opts apiv1.Options) (apiv1.KeyManager, error) {
 		name := filepath.Base(os.Args[0])
-		return nil, errors.Errorf("unsupported kms type 'yubikey': %s is compiled without cgo support", name)
+		return nil, errors.Errorf("unsupported kms type 'yubikey': %s is compiled without CGO or YubiKey support", name)
```
hslatman, line 18: I'm okay with `CGO`, but the Go docs use `cgo`

kms/capi/no_capi.go
```diff
@@ -0,0 +1,20 @@
...
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"go.step.sm/crypto/kms/apiv1"
+)
+
+func init() {
+	apiv1.Register(apiv1.CAPIKMS, func(ctx context.Context, opts apiv1.Options) (apiv1.KeyManager, error) {
+		name := filepath.Base(os.Args[0])
+		return nil, errors.Errorf("unsupported kms type 'capi': %s is compiled without Microsoft CryptoAPI support", name)
```
hslatman, line 18: Does this need the call stack? I think `fmt.Errorf` is sufficient for those errors and removes unwanted dependency on `github.com/pkg/errors`

maraino: That package was used in all the rest of the kms code, so I didn't change it.

kms/yubikey/yubikey_no_cgo.go
maraino: Fixed with 5744559

kms/capi/no_capi.go
hslatman: Then we're never going to get rid of it 😥

hslatman (approved):

kms/capi/no_capi.go
maraino: If we want to get rid of it we'll probably should do it in PRs that cover one or more packages. At that time we will need to decide a few things, should we use more typed errors? Should we the `errors.Join`? Do we want to keep a stack trace?

PR #177, opened 2023-02-14
TPM utilities package

hslatman:


tpm/ak.go
```diff
@@ -0,0 +1,204 @@
...
+	now := time.Now()
+
+	if name == "" {
+		// TODO: decouple the TPM key name from the name recorded in the storage? This might
+		// make it easier to work with the key names as a user; the TPM key name would be abstracted
+		// away. The key name in the storage can be different from the key stored with the key (which,
+		// to be far, isn't even used on Linux TPMs)
+		nameHex := make([]byte, 5)
+		if n, err := rand.Read(nameHex); err != nil || n != len(nameHex) {
+			return result, fmt.Errorf("rand.Read() failed with %d/%d bytes read and error: %w", n, len(nameHex), err)
```
azazeal, line 44: The number of bytes read aren't all that important here. A much simpler way to express this failure would be something like: 

```go
fmt.Errorf("failed reading from CSPRNG: %w", err")
```

tpm/ak.go
```diff
@@ -0,0 +1,204 @@
...
+
+	at, err := attest.OpenTPM(t.attestConfig)
+	if err != nil {
+		return result, fmt.Errorf("failed opening TPM: %w", err)
+	}
+	defer at.Close()
+
+	now := time.Now()
+
+	if name == "" {
```
azazeal, line 37: I think you should extract this name generation code into its own function (as it's used in a couple of places).

tpm/ak.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	storedAK := &storage.AK{
+		Name:      name,
+		Data:      data,
+		CreatedAt: now,
+	}
+
+	if err := t.store.AddAK(storedAK); err != nil {
+		return result, err
```
azazeal, line 72: You should wrap the error, `failed adding AK: %w`

tpm/ak.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	if err := t.store.Persist(); err != nil {
+		return result, err
```
azazeal, line 76: Wrap the error: `failed persisting: %w` or something along those lines

tpm/ak.go
```diff
@@ -0,0 +1,204 @@
...
+		if n, err := rand.Read(nameHex); err != nil || n != len(nameHex) {
+			return result, fmt.Errorf("rand.Read() failed with %d/%d bytes read and error: %w", n, len(nameHex), err)
+		}
+		name = fmt.Sprintf("%x", nameHex)
+	}
+
+	prefixedName := fmt.Sprintf("ak-%s", name)
+
+	akConfig := attest.AKConfig{
+		Name: prefixedName,
```
azazeal, line 52: since `prefixedName` is only referenced here you can just as easily do 

```go
Name: fmt.Sprintf("ak-%s", name),
```

tpm/ak.go
```diff
@@ -0,0 +1,204 @@
...
+	"crypto/rand"
+	"fmt"
+	"time"
+
+	"github.com/google/go-attestation/attest"
+
+	"go.step.sm/crypto/tpm/storage"
+)
+
+type AK struct {
```
azazeal, line 14: There's lots of places where you `return result, err`. Maybe turn result into a pointer and start returning nil?

tpm/ak.go
```diff
@@ -0,0 +1,204 @@
...
+
+func (t *TPM) ListAKs(ctx context.Context) ([]AK, error) {
+	if err := t.Open(ctx); err != nil {
+		return nil, fmt.Errorf("failed opening TPM: %w", err)
+	}
+	defer t.Close(ctx, false)
+
+	aks, err := t.store.ListAKs()
+	if err != nil {
+		return nil, fmt.Errorf("error listing AKs: %w", err)
```
azazeal, line 105: `failed listing` more consistent

tpm/ak.go
```diff
@@ -0,0 +1,204 @@
...
+func (t *TPM) GetAK(ctx context.Context, name string) (AK, error) {
+	result := AK{}
+	if err := t.Open(ctx); err != nil {
+		return result, fmt.Errorf("failed opening TPM: %w", err)
+	}
+	defer t.Close(ctx, false)
+
+	ak, err := t.store.GetAK(name)
+	if err != nil {
+		return result, fmt.Errorf("error getting AK %q: %w", name, err)
```
azazeal, line 91: `failed` more consistent

tpm/ak.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	return AK{Name: ak.Name, Data: ak.Data, CreatedAt: ak.CreatedAt, tpm: t}, nil
+}
+
+func (t *TPM) ListAKs(ctx context.Context) ([]AK, error) {
+	if err := t.Open(ctx); err != nil {
+		return nil, fmt.Errorf("failed opening TPM: %w", err)
+	}
+	defer t.Close(ctx, false)
```
azazeal, line 101: can close return an error? if so, you should return it.

I know it sounds counter-intuitive but this may very well be more important than the result you retrieve some times.

tpm/context.go
```diff
@@ -0,0 +1,13 @@
+package tpm
```
azazeal, line 1: this file is :fire: - i'm luvin' it

azazeal (commented):
Few initial thoughts; good stuff overall

tpm/storage/dirstore.go
```diff
@@ -0,0 +1,231 @@
...
+	ext := filepath.Ext(pathKey.FileName)
+	if ext != tpmExtension { // skipping
+		return ""
+	}
+	filename := pathKey.FileName[:len(pathKey.FileName)-len(tpmExtension)]
+	p := filepath.Join(filepath.Join(pathKey.Path...), filename)
+	if len(pathKey.Path) > 0 && pathKey.Path[0] == "" { // absolute path at "/"
+		p = filepath.Join(string(filepath.Separator), p)
+	}
+	return p
```
maraino, line 41: Other filepath methods might be useful here, Abs, IsAbs, Clean.

tpm/storage/types.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (ak *AK) UnmarshalJSON(data []byte) error {
+	sak := &serializedAK{}
+	if err := json.Unmarshal(data, sak); err != nil {
+		return fmt.Errorf("failed unmarshaling serialized AK: %w", err)
+	}
+
+	ak.Name = sak.Name
+	ak.Data = sak.Data
+	ak.CreatedAt = sak.CreatedAt
```
maraino, line 45: Do we need to verify the Type == typeAK?

tpm/storage/types.go
```diff
@@ -0,0 +1,152 @@
...
+func (key *Key) UnmarshalJSON(data []byte) error {
+	sk := &serializedKey{}
+	if err := json.Unmarshal(data, sk); err != nil {
+		return fmt.Errorf("failed unmarshaling serialized key: %w", err)
+	}
+
+	key.Name = sk.Name
+	key.Data = sk.Data
+	key.AttestedBy = sk.AttestedBy
+	key.CreatedAt = sk.CreatedAt
```
maraino, line 100: Do we need to verify the type?

maraino (commented):
I went through the code, not the tests, and added some comments. One thing I've missed is a little bit more documentation on types and methods. Perhaps not on every method, but we should document at least the types.

tpm/storage/types.go
hslatman: Nice one. Added in 1d8a8b434257ae7d1338ba30b54f9bf8abb93ca4.

tpm/storage/types.go
hslatman: Added in 1d8a8b434257ae7d1338ba30b54f9bf8abb93ca4

tpm/skae/extension.go
```diff
@@ -0,0 +1,196 @@
...
+
+	result := pkix.Extension{
+		Id:       oidSubjectKeyAttestationEvidence,
+		Critical: false, // non standard extension; don't break clients
+		Value:    skaeExtensionBytes,
+	}
+
+	q.Q(result)
+	q.Q(string(result.Value))
+	q.Q(base64.StdEncoding.EncodeToString(result.Value))
```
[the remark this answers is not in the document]
hslatman: Yes, will do. I'm not using this actively yet; I need to verify that the contents are OK, but haven't found examples so far.

tpm/signer.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	defer t.Close(ctx)
+
+	key, err := t.store.GetKey(name)
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			return nil, fmt.Errorf("failed getting signer for key %q: %w", name, ErrNotFound)
+		}
+		return nil, err
+	}
```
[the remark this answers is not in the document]
hslatman: Added the check here too. Due to the nature of the package, in `Sign` the check is also still performed, because the key is unloaded from the TPM in the `defer`. So currently every call to `Sign` reloads the key. We could add a method that loads the key once, then can be used for multiple sign operations, with the caller of the method being responsible for unloading the key.

tpm/key.go
```diff
@@ -0,0 +1,410 @@
...
+}
+
+type AttestKeyConfig struct {
+	// Algorithm to be used, either RSA or ECDSA.
+	Algorithm string
+	// Size is used to specify the bit size of the key or elliptic curve. For
+	// example, '256' is used to specify curve P-256.
+	Size int
+
+	QualifyingData []byte
```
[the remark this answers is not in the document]
hslatman: Done in 1d8a8b434257ae7d1338ba30b54f9bf8abb93ca4.

tpm/ak.go
```diff
@@ -0,0 +1,374 @@
...
+	"time"
+
+	"github.com/google/go-attestation/attest"
+	x509ext "github.com/google/go-attestation/x509"
+
+	"go.step.sm/crypto/tpm/storage"
+)
+
+// AK models a TPM 2.0 Attestation Key.
+type AK struct {
```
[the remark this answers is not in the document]
hslatman: Done in 1d8a8b434257ae7d1338ba30b54f9bf8abb93ca4. Hope my intention with the package is clear 🙂

tpm/internal/key/key.go
```diff
@@ -0,0 +1,267 @@
...
+				XRaw: make([]byte, 48),
+				YRaw: make([]byte, 48),
+			}
+		case 521:
+			tmpl.NameAlg = tpm2.AlgSHA512
+			tmpl.ECCParameters.Sign.Hash = tpm2.AlgSHA512
+			tmpl.ECCParameters.CurveID = tpm2.CurveNISTP521
+			tmpl.ECCParameters.Point = tpm2.ECPoint{
+				XRaw: make([]byte, 65),
+				YRaw: make([]byte, 65),
```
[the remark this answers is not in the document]
hslatman: This is in code that I copied from [go-attestation](https://github.com/google/go-attestation/blob/270ecbab1f21f1a2f10d30553a5666d2ae01fbbc/attest/wrapped_tpm20.go#L247-L292). I noticed there's an open issue for this there: https://github.com/google/go-attestation/issues/303. I'll see if I can somehow find an example of a TPM with P521.

tpm/internal/key/key.go
```diff
@@ -0,0 +1,267 @@
...
+	Name string
+	// Blob represents the key material for KeyEncodingEncrypted keys. This
+	// is only used on Linux.
+	Blob []byte `json:"KeyBlob"`
+}
+
+// Serialize represents the key in a persistent format which may be
+// loaded at a later time using deserializeKey().
+func (k *serializedKey) Serialize() ([]byte, error) {
+	return json.Marshal(k)
```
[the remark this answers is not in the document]
hslatman: This part is also copied from https://github.com/google/go-attestation. To keep the data blob for non-attested keys compatible with attested keys, I propose to keep it like this for now.

tpm/ek.go
```diff
@@ -0,0 +1,181 @@
...
+func (ek *EK) MarshalJSON() ([]byte, error) {
+	var der []byte
+	if ek.certificate != nil {
+		der = ek.certificate.Raw
+	}
+	o := struct {
+		Type string `json:"type"`
+		DER  []byte `json:"der,omitempty"`
+		URL  string `json:"url,omitempty"`
+	}{
```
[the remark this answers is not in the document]
hslatman: Made them all anonymous in 1d8a8b434257ae7d1338ba30b54f9bf8abb93ca4.

tpm/ek.go
```diff
@@ -0,0 +1,181 @@
...
+	"net/url"
+	"path"
+	"strings"
+
+	"github.com/google/go-attestation/attest"
+)
+
+type EK struct {
+	public         crypto.PublicKey
+	certificate    *x509.Certificate
```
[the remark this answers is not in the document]
hslatman: Not as far as I know. The underlying library also doesn't return a chain.

tpm/ak.go
```diff
@@ -0,0 +1,374 @@
...
+		if ext.Id.Equal(oidSubjectAlternativeName) {
+			sanExtension = ext
+		}
+	}
+
+	if sanExtension.Value == nil {
+		return false
+	}
+
+	san, err := x509ext.ParseSubjectAltName(sanExtension) // TODO(hs): move to a package under our control?
```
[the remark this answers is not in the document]
hslatman: Will do.

tpm/ak.go
```diff
@@ -0,0 +1,374 @@
...
+	chain := ak.chain
+	if len(chain) == 0 {
+		return false
+	}
+	akCert := chain[0]
+
+	var sanExtension pkix.Extension
+	for _, ext := range akCert.Extensions {
+		if ext.Id.Equal(oidSubjectAlternativeName) {
+			sanExtension = ext
```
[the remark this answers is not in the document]
hslatman: Fixed in 1d8a8b434257ae7d1338ba30b54f9bf8abb93ca4

tpm/ak.go
```diff
@@ -0,0 +1,374 @@
...
+}
+
+func (ak *AK) SetCertificateChain(ctx context.Context, chain []*x509.Certificate) error {
+	if err := ak.tpm.Open(ctx); err != nil {
+		return fmt.Errorf("failed opening TPM: %w", err)
+	}
+	defer ak.tpm.Close(ctx)
+
+	// TODO(hs): perform validation, such as check if the chain includes leaf for the
+	// AK public key?
```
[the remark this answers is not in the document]
hslatman: Done in 1d8a8b434257ae7d1338ba30b54f9bf8abb93ca4. 

I'm not entirely happy with the approach yet. I think I'll (de)serialize the AK public key, so that it can be more easily reloaded instead of having to interact with the TPM for this.

tpm/ek.go
```diff
@@ -0,0 +1,181 @@
...
+			if r.StatusCode != http.StatusOK {
+				return nil, fmt.Errorf("http request to %q failed with status %d", ekURL, r.StatusCode)
+			}
+
+			var c intelEKCertResponse
+			if err := json.NewDecoder(r.Body).Decode(&c); err != nil {
+				return nil, fmt.Errorf("error decoding EK certificate response: %w", err)
+			}
+
+			cb, err := base64.URLEncoding.DecodeString(c.Certificate)
```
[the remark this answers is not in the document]
hslatman: That means my current logic is correct, right? I'm not entirely sure why this URL doesn't seem to be made to contain the right URL by the [go-attestation code](https://github.com/google/go-attestation/blob/270ecbab1f21f1a2f10d30553a5666d2ae01fbbc/attest/tpm.go#L212-L223) yet. Their [test](https://github.com/google/go-attestation/blob/270ecbab1f21f1a2f10d30553a5666d2ae01fbbc/attest/tpm_test.go#L47-L53) seems to expect `=`, which is weird, I think? We can't do `WithPadding("%3D")`, as it takes a rune; not a string. I think it's best to fix this in go-attestation, but the current code works as a workaround.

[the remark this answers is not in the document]
hslatman: Also see the last comment in https://community.intel.com/t5/Intel-NUCs/TPM-EK-Certificate-Retrieval/td-p/1345477/page/2. They may have changed something. Using that EK URL, but with `=` replaced by `%3D`, it works.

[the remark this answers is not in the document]
maraino: I meant that the .certificate also ends with %3D, not just the URL

tpm/internal/key/key.go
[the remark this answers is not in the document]
maraino: There's a TPM emulator that supports those keys. There's a comment that links to https://github.com/stefanberger/swtpm.

tpm/ek.go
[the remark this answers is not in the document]
hslatman: Ah, interesting. Have to replace it in the response then too, I suppose. 

I've opened a PR: https://github.com/google/go-attestation/pull/310.

tpm/ak.go
hslatman: Made the method private and am now catching the error on `close` in the defer. Errors originating from the main logic have priority, so I check if no error was set, and then override it with the one from `close` (if set).

tpm/ek.go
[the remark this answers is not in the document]
hslatman: I've pushed some changes recently that allow us to do manufacturer specific logic. For Intel an example certificate is available for testing.

tpm/internal/key/key.go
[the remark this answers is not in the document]
hslatman: Made this 66 in the copied code.

tpm/ek.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+// prepareEKCertificateURL prepares the URL from which an EK can be downloaded.
+// It parses the provided ekURL. If the TPM manufacturer is Intel, we patch the URL to
+// have the right format. This should become redundant when https://github.com/google/go-attestation/pull/310
+// is merged.
+func (t *TPM) prepareEKCertifiateURL(ctx context.Context, ekURL string) (*url.URL, error) {
+	var u *url.URL
+	var err error
+	u, err = url.Parse(ekURL)
```
azazeal, line 203: you don't need the `var`s above this, yes?

tpm/ek.go
```diff
@@ -0,0 +1,318 @@
...
+		u, err = url.Parse(s)
+		if err != nil {
+			return nil, fmt.Errorf("failed parsing EK certificate URL %q: %w", s, err)
+		}
+	}
+
+	return u, nil
+}
+
+func (t *TPM) downloadEKCertifiate(ctx context.Context, ekURL *url.URL) (*x509.Certificate, error) {
```
azazeal, line 237: typoman is better than superman!

tpm/errors.go
```diff
@@ -0,0 +1,9 @@
+package tpm
+
+import "errors"
+
+// ErrNotFound is returned when a Key or AK cannot be found
+var ErrNotFound = errors.New("not found")
```
azazeal, line 6: maybe prefix these errors with `tpm: ` like, `tpm: not found`.

tpm/info.go
```diff
@@ -0,0 +1,159 @@
...
+// MarshalJSON marshals the version into JSON.
+func (v Version) MarshalJSON() ([]byte, error) {
+	var s string
+	switch v {
+	case Version(attest.TPMVersion12):
+		s = "1.2"
+	case Version(attest.TPMVersion20):
+		s = "2.0"
+	default:
+		s = "unknown"
```
azazeal, line 48: `s = fmt.Sprintf("unknown (%d)", v)` or in hex. whatever makes more sense to you.

tpm/info.go
```diff
@@ -0,0 +1,159 @@
...
+	case Interface(attest.TPMInterfaceDirect):
+		return "direct"
+	case Interface(attest.TPMInterfaceKernelManaged):
+		return "kernel-managed"
+	case Interface(attest.TPMInterfaceDaemonManaged):
+		return "daemon-managed"
+	case Interface(attest.TPMInterfaceCommandChannel):
+		return "command-channel"
+	default:
+		return "unknown"
```
azazeal, line 69: maybe also include some indication as to the actual value, like above.

tpm/info.go
```diff
@@ -0,0 +1,159 @@
...
+	Name  string          `json:"name"`
+	ASCII string          `json:"ascii"`
+	Hex   string          `json:"hex"`
+}
+
+// String returns a textual representation of the TPM
+// manufacturer. An example looks like this:
+//
+//	ST Microelectronics (STM, 53544D20, 1398033696)
+func (m Manufacturer) String() string {
```
azazeal, line 108: this should maybe a pointer receiver.

tpm/info.go
```diff
@@ -0,0 +1,159 @@
...
+		ID:    id,
+		Hex:   hexa,
+	}
+}
+
+// Info returns info about the TPM. The info doesn't change, so
+// it's cached after the first lookup.
+func (t *TPM) Info(ctx context.Context) (*Info, error) {
+	if t.info == nil {
+		var err error
```
azazeal, line 129: maybe refactor so that `err` is a named return value, like earlier.

tpm/internal/key/key.go
```diff
@@ -0,0 +1,267 @@
+//nolint:errorlint,revive // copied code from github.com/google/go-attestation
```
azazeal, line 1: since this is copied code add this to the header:

`CODE COPIED FROM ...; DO NOT EDIT!`

tpm/key.go
```diff
@@ -0,0 +1,505 @@
...
+
+	// TODO(hs): add akName and key name to this struct?
+}
+
+// CreateKey creates a new Key identified by `name`. If no name is  provided,
+// a random 10 character name is generated. If a Key with the same name exists,
+// `ErrExists` is returned. The Key won't be attested by an AK.
+func (t *TPM) CreateKey(ctx context.Context, name string, config CreateKeyConfig) (*Key, error) {
+	var err error
```
azazeal, line 147: a named return, perhaps, if it'll be used with `defer`

tpm/key.go
```diff
@@ -0,0 +1,505 @@
...
+
+	return key, nil
+}
+
+// AttestKey creates a new Key identified by `name` and attested by the AK
+// identified by `akName`. If no name is  provided, a random 10 character
+// name is generated. If a Key with the same name exists, `ErrExists` is
+// returned.
+func (t *TPM) AttestKey(ctx context.Context, akName, name string, config AttestKeyConfig) (*Key, error) {
+	var err error
```
azazeal, line 198: named return :)

tpm/key.go
```diff
@@ -0,0 +1,505 @@
...
+	if err = t.open(ctx); err != nil {
+		return nil, fmt.Errorf("failed opening TPM: %w", err)
+	}
+	defer func() {
+		if tempErr := t.close(ctx); tempErr != nil && err != nil {
+			err = tempErr
+		}
+	}()
+
+	now := time.Now().UTC()
```
azazeal, line 157: Please do check around the codebase to make sure that all date handling ends up being either all UTC or all Local and not mixed, please.

tpm/key.go
```diff
@@ -0,0 +1,505 @@
...
+	return
+}
+
+// Blobs returns a container for the private and public key blobs.
+// The resulting blobs are compatible with tpm2-tools, so can be used
+// like this (after having been written to key.priv and key.pub):
+//
+//	tpm2_load -C 0x81000001 -u key.pub -r key.priv -c key.ctx
+func (k *Key) Blobs(ctx context.Context) (*Blobs, error) {
+	if k.blobs == nil {
```
azazeal, line 412: maybe reverse the clause and deindent?

tpm/manufacturer/manufacturers.go
```diff
@@ -0,0 +1,102 @@
...
+var (
+	manufacturerByASCII map[string]string
+	validChars          *regexp.Regexp
+)
+
+// ID models a TPM Manufacturer (or Vendor) ID.
+type ID uint32
+
+func (id ID) MarshalJSON() ([]byte, error) {
+	return json.Marshal(strconv.FormatUint(uint64(id), 10))
```
azazeal, line 21: Maybe add a note that this should be formatted as a string.

tpm/simulator/simulator_disabled.go
```diff
@@ -0,0 +1,39 @@
...
+func (s *Simulator) Close() error {
+	return errors.New("cannot close: no simulator available")
+}
+
+func (s *Simulator) MeasurementLog() ([]byte, error) {
+	return nil, errors.New("cannot get measurement log: no simulator available")
+}
+
+func (s *Simulator) Read(p []byte) (n int, err error) {
+	return 0, errors.New("cannot read: no simulator available")
```
azazeal, line 32: you don't need named returns here.

tpm/storage/dirstore.go
```diff
@@ -0,0 +1,236 @@
...
+func advancedTransform(key string) *diskv.PathKey {
+	path := strings.Split(key, "/")
+	last := len(path) - 1
+	return &diskv.PathKey{
+		Path:     path[:last],
+		FileName: path[last] + tpmExtension,
+	}
+}
+
+func inverseTransform(pathKey *diskv.PathKey) (key string) {
```
azazeal, line 33: you have a named return here, so refactor to either keep it or remove it.

tpm/tpm.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// clean up the go-tpm rwc
+	if t.rwc != nil {
+		err := t.rwc.Close()
+		t.rwc = nil
+		return fmt.Errorf("failed closing rwc: %w", err)
+	}
+
+	// mark the TPM as ready to be used again
+	t.lock.Unlock()
```
maraino, line 180: we are not unlocking this if it fails before.

tpm/tpm.go
```diff
@@ -0,0 +1,183 @@
...
+// concurrent use.
+func (t *TPM) open(ctx context.Context) error {
+	// prevent opening the TPM multiple times if Open is called
+	// within the package multiple times.
+	if isInternalCall(ctx) {
+		return nil
+	}
+
+	// lock the TPM instance; it's in use now
+	t.lock.Lock()
```
maraino, line 99: perhaps we should unlock on errors.

maraino (changes requested):
Some extra comments, a lot of defers with ignored errors.

tpm/ak.go
```diff
@@ -136,7 +135,7 @@ func (t *TPM) CreateAK(ctx context.Context, name string) (*AK, error) {
 	}
 
 	if _, err := t.store.GetAK(name); err == nil {
-		return nil, fmt.Errorf("failed creating AK %q: %w", name, ErrExists)
+		return nil, fmt.Errorf("failed getting AK %q: %w", name, ErrExists)
```
maraino, line 138: The error is creating the key, not getting it. It cannot create the key because it exists. Getting the key does not fail.

tpm/errors.go
hslatman: These errors are wrapped internally. The resulting error would be something like `failed getting AK "name": not found`. Prefixing these two errors won't look nice. But I could prefix the top level errors returned from the `tpm` package.

tpm/ak.go
```diff
@@ -0,0 +1,507 @@
...
+	}
+
+	return nil
+}
+
+// AttestationParameters returns information about the AK, typically used to
+// generate a credential activation challenge.
+func (ak *AK) AttestationParameters(ctx context.Context) (params attest.AttestationParameters, err error) {
+	if ak.attestParams == nil {
+		var err error
```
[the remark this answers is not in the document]
hslatman: I have rewritten this to rely on named `err error` returns and a `closeTPM` function executed with `defer`. It will keep the `err` from main-line code. Only when closing the TPM returns an error and main-line code didn't, it will override the error returned. Changed this for all `defer ...` calls in this PR.

tpm/skae/extension.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return pkix.Extension{}, fmt.Errorf("error marshaling issuer: %w", err)
+	}
+
+	skaeExtension := asn1SKAE{
+		TCGSpecVersion:         asn1TCGSpecVersion{Major: 2, Minor: 0},
+		KeyAttestationEvidence: asn1KeyAttestationEvidence{},
+	}
+
+	attestationEvidence := asn1AttestationEvidence{
```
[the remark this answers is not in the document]
hslatman: This extension is currently not being used. I've made a note of adding this later.

tpm/ak.go
```diff
@@ -0,0 +1,507 @@
...
+		return fmt.Errorf("failed getting keys attested by AK %q: %w", name, err)
+	}
+
+	if len(keys) > 0 {
+		return fmt.Errorf("cannot delete AK %q before deleting keys that were attested by it", name)
+	}
+
+	if err := t.attestTPM.DeleteKey(ak.Data); err != nil { // TODO: we could add a DeleteAK to go-attestation; under the hood it's loaded the same as a key though.
+		return fmt.Errorf("failed deleting AK %q: %w", name, err)
+	}
```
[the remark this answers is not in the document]
hslatman: On Linux the call doesn't do anything currently, because the TPM doesn't store the key. On Windows we might get an error like `0x8028000D: "TPM_E_KEYNOTFOUND"`, when doing `nCryptDeleteKey`. The full error would look something like this: `nCryptDeleteKey returned 8028000D: tpm or subsystem failure: TPM_E_KEYNOTFOUND`. We can only rely on checking the string, currently. 

Are you OK with a follow-up PR for this? I've made a note of this.

tpm/tpm.go
hslatman: I added a `defer` with a check to do so.

tpm/ak.go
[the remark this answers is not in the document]
maraino: 👍

maraino (approved):
lgtm, I've added a suggestion.

tpm/ak.go
```diff
@@ -0,0 +1,469 @@
...
+}
+
+// CreateAK creates and stores a new AK identified by `name`.
+// If no name is  provided, a random 10 character name is generated.
+// If an AK with the same name exists, `ErrExists` is returned.
+func (t *TPM) CreateAK(ctx context.Context, name string) (ak *AK, err error) {
+	if err = t.open(ctx); err != nil {
+		return nil, fmt.Errorf("failed opening TPM: %w", err)
+	}
+	defer closeTPM(ctx, t, &err)
```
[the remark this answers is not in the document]
hslatman: An error from closing the TPM will only be returned when `&err` hasn't been set to an error.

tpm/ek.go
```diff
@@ -0,0 +1,323 @@
...
+	switch t := p.(type) {
+	case *rsa.PublicKey:
+		return fmt.Sprintf("RSA %d", t.Size()*8)
+	case *ecdsa.PublicKey:
+		switch size := t.Curve.Params().BitSize; size {
+		case 256, 384, 521:
+			return fmt.Sprintf("ECDSA P-%d", size)
+		default:
+			return fmt.Sprintf("unexpected ECDSA size: %d", size)
+		}
```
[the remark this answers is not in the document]
hslatman: That's simpler; thx 🙂

maraino (approved):

PR #178, opened 2023-02-17
TPM simulator

hslatman:
The TPM simulator from https://github.com/google/go-tpm-tools is used to test TPM interactions.


maraino:
@hslatman Do we really need to vendor?

hslatman:
Strictly speaking, I don't think it's required. I probably added it out of habit. I don't think it's bad necessary; it's an extra way of knowing what code your build is using, apart from the hashes and the Go module proxy. Given that `crypto` is more of a library, maybe I should remove it again. 

There's one interesting thing about why it's useful for this specific PR though: the https://github.com/google/go-tpm-tools repository includes the C source code for the TPM simulator. That code isn't `go get`'d automatically; Go ignores it. By having `github.com/google/go-tpm-tools` in the vendor dir, and after manual copying of the C source code into that dir, the simulator can be used successfully. Otherwise it'll be missing the required header files. The problem is explained [here](https://github.com/google/go-tpm-tools/issues/40) and [here](https://github.com/google/go-tpm-tools/issues/274).

Do you have a suggestion on how to do this without having the vendor dir? Including it as a submodule and partly rewriting the search/import paths might work?

maraino:
> Do you have a suggestion on how to do this without having the vendor dir? Including it as a submodule and partly rewriting the search/import paths might work? Only downloading the TPM source dir in GH actions might also be an option, but then it's not as easy to run locally.

Perhaps we can copy the full `simulator` folder, change the necessary imports, and keep just that. We might want to add specific tags to enable the simulator too.

hslatman:
Sounds OK. Will try that. Simulator is only used in tests that have the cgo tag. Testing with cgo disabled skips those.
________________________________

hslatman:
Added `tpmsimulator` build tag. Not providing it skips the simulator tests. It also requires `CGO_ENABLED=1` if specified.

tpm/simulator/internal/simulator.go
```diff
@@ -0,0 +1,158 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+// Package simulator provides a go interface to the Microsoft TPM2 simulator.
+//
+//nolint:gosec // copied package from github.com/google/go-tpm-tools/simulator; intentionally using math.rand for seed
+package internal
```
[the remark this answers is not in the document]
maraino: I guess it is easier to copy the full directory to an internal package and deal with it. Up to you.

tpm/tpm.go
```diff
@@ -64,6 +71,36 @@ func (t *TPM) Open(ctx context.Context) error {
...
+			}
+			t.rwc = rwc
+		} else {
+			at, err := attest.OpenTPM(t.attestConfig)
+			if err != nil {
+				return fmt.Errorf("failed opening TPM: %w", err)
+			}
+			t.attestTPM = at
+		}
+	}
```
maraino, line 102: Ok, optionally an `open` method in `go:build tpmsimulator` and another in `go:build !tpmsimulator`.

maraino: Ideally you should be able to enable the simulator from a third-party package too just using the tags. This can also be handled in the open package too.

maraino (commented):
Generally ok, although there's a bunch of code that is not part of this review.

tpm/simulator/internal/internal/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package internal provides low-level bindings to the Microsoft TPM2 simulator.
+//
+// When using CGO, this package compiles the simulator's C code and links
+// against the system OpenSSL library. Without CGO, this package just provides
+// stubs which always return failure. This allows the simulator package to be
+// built when cross compiling go-tpm-tools (which is incompatible with CGO).
+package internal
```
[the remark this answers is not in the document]
hslatman: Yea, I agree it's not super nice, but this is the easiest way to just copy/paste the entire directory with minimal changes to the original code and no changes to import paths. It's all internal, and not meant to be used outside of the tests for this package (at the moment), so I think it's OK to leave it as is.

tpm/tpm.go
hslatman: At the moment I do not intend the C TPM simulator in this repo to be used externally. In fact, people trying to use it would have the same issue `go get`-ing the dependency as we had in the first place. Or are you thinking of different use cases?

tpm/ek.go
```diff
@@ -18,40 +18,52 @@ import (
...
 		}
 	}
-	o := out{
-		Type: fmt.Sprintf("%T", ek.Public), // TODO: proper description string; on EK struct
+	o := struct {
+		Type string `json:"type"`
+		PEM  string `json:"pem,omitempty"`
+		URL  string `json:"url,omitempty"`
+	}{
+		Type: fmt.Sprintf("%T", ek.public), // TODO: proper description string; on EK struct
```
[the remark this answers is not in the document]
hslatman: Yes, absolutely.

tpm/ek.go
```diff
@@ -18,40 +18,52 @@ import (
...
+	if ek.certificate != nil {
 		if pemString, err = ek.PEM(); err != nil {
 			return nil, err
 		}
 	}
-	o := out{
-		Type: fmt.Sprintf("%T", ek.Public), // TODO: proper description string; on EK struct
+	o := struct {
+		Type string `json:"type"`
+		PEM  string `json:"pem,omitempty"`
```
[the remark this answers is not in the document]
hslatman: I'll see if it makes sense to provide DER by default and keep providing the utilities for the PEM format. I think in `step-tpm-plugin`, the PEM output is more useful, but for direct outputs, having the raw bytes may be easier.

tpm/tpm.go
maraino: This code works, and I can see the .c and .h files in my `~/go/pkg/mod/...`. I don't know if more complex stuff won't, but if this library provides useful stuff that we can use for testing or while developing on a macOS, I would say we should allow it.

```go
package main

import (
	"fmt"
	"log"

	"github.com/google/go-tpm-tools/simulator"
	"github.com/google/go-tpm/tpm2"
)

func main() {
	sim, err := simulator.Get()
	if err != nil {
		log.Fatal(err)
	}
	defer sim.Close()

	b, err := tpm2.GetRandom(sim, 32)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("%x\n", b)
}
```

hslatman: I moved back to where I started. Just removed the copy for now and rely on the import directly, but the tags are still present. Third party packages should now be able to do `simulator.New()` (from this package). If they vendor their dependencies, it won't work.

PR #185, opened 2023-03-03
Add support for multiple yubikeys at the same time

maraino:
### Description

This PR allows supporting multiple YubiKeys at the same time. We will use the serial number in the attestation certificate to differentiate those. If a serial is not provided, we will open the first card on the list of cards.

### DIFF
--- a/kms/yubikey/yubikey_test.go
+++ b/kms/yubikey/yubikey_test.go
@@ -14,6 +14,7 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"encoding/asn1"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


kms/yubikey/yubikey_test.go
```diff
@@ -988,3 +1017,64 @@ func TestYubiKey_Close(t *testing.T) {
...
+
+	failAttest := newStubPivKey(t, RSA)
+	delete(failAttest.attestMap, piv.SlotAuthentication)
+
+	failParse := newStubPivKey(t, ECDSA)
+	serialNumer, err := asn1.Marshal("112233")
+	if err != nil {
+		t.Fatal(err)
+	}
+	attCertParse, err := failParse.attestCA.Sign(&x509.Certificate{
```
areed, line 1032: Why isn't this going to parse successfully?

maraino: I am adding the serial number extension with a string (ASN.1 PrintableString Type `asn1.Marshal("112233")` while the extension expects an integer. If you look at the function `getSerialNumber()`, we try to unmarshal to an integer, and that test case fails there.

areed (approved):

PR #204, opened 2023-03-24
Add permanent identifier in the CreateAttestation response

maraino:
This commit adds an optional permanent identifier in the CreateAttestation response. In the YubiKey KMS, the permanent identifier is the serial number in the attestation certificate.

This PR also files all linter errors.

### DIFF
--- a/kms/yubikey/yubikey.go
+++ b/kms/yubikey/yubikey.go
@@ -475,7 +475,7 @@ func getPolicies(req *apiv1.CreateKeyRequest) (piv.PINPolicy, piv.TouchPolicy) {
 }
 
 // getSerialNumber returns the serial number from an attestation certificate. It
-// will return an empty string if it the serial number extension does not exists
+// will return an empty string if the serial number extension does not exist
 // or if it is malformed.
 func getSerialNumber(cert *x509.Certificate) string {
 	for _, ext := range cert.Extensions {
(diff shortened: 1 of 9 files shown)


kms/yubikey/yubikey.go
```diff
@@ -471,22 +474,19 @@ func getPolicies(req *apiv1.CreateKeyRequest) (piv.PINPolicy, piv.TouchPolicy) {
...
-// getSerialNumber gets an attestation certificate on the given key and returns
-// the serial number on it.
-func getSerialNumber(yk pivKey) (string, error) {
-	cert, err := yk.Attest(piv.SlotAuthentication)
-	if err != nil {
-		return "", err
-	}
+// getSerialNumber returns the serial number from an attestation certificate. It
+// will return an empty string if it the serial number extension does not exists
+// or if it is malformed.
```
hslatman, line 479:
```suggestion
// getSerialNumber returns the serial number from an attestation certificate. It
// will return an empty string if the serial number extension does not exist
// or if it is malformed.
```

hslatman (commented):
Looks OK for the most part. Only reserve I have is if this should be called `PermanentIdentifier`, or something slightly more generic? While it's certainly used like a PermanentIdentifier would be used, this isn't currently coming from an actual PermanentIdentifier SAN, nor do we encode the Assigner with it (although the Issuer of the attestation cert can be used for that in lieu of the Assigner). It's a great name for it is and does, though. Some alternatives would be PersistentIdentifier and DurableIdentifier, but I still like PermanentIdentifier better. Identifier feels to generic and doesn't make clear that it can't/won't change, so also not great.

hslatman (approved):

PR #207, opened 2023-03-28
Fix SSH certificate fingerprint encoding

hslatman:
Before the change the fingerprint for an SSH certificate would
include the bytes that encode the certificate. An SSH fingerprint
should only be based on its public key.

This commit checks if the SSH key is actually an SSH certificate
and will fingerprint just the SSH certificate public key instead
of the entire certificate contents.

### DIFF
--- a/sshutil/fingerprint.go
+++ b/sshutil/fingerprint.go
@@ -62,10 +62,41 @@ func EncodedFingerprint(pub ssh.PublicKey, encoding FingerprintEncoding) string 
 // OpenSSH and returns a public key fingerprint in the following format:
 //
 //	<size> SHA256:<base64-raw-fingerprint> <comment> (<type)
+//
+// If the input is an SSH certificate, its public key will be extracted and
+// taken as input for the fingerprint.
 func FormatFingerprint(in []byte, encoding FingerprintEncoding) (string, error) {
+	return formatFingerprint(in, encoding, false)
+}
+
+// FormatCertificateFingerprint parses an SSH certificate as used by
+// OpenSSH and returns a public key fingerprint in the following format:
+//
+//	<size> SHA256:<base64-raw-fingerprint> <comment> (<type)
+//
+// If the input is not an SSH certificate, an error will be returned.
+func FormatCertificateFingerprint(in []byte, encoding FingerprintEncoding) (string, error) {
+	return formatFingerprint(in, encoding, true)
+}
+
+// formatFingerprint parses a public key from an authorized_keys file or an
+// SSH certificate as used by OpenSSH and returns a public key fingerprint
+// in the following format:
+//
+//	<size> SHA256:<base64-raw-fingerprint> <comment> (<type)
+//
+// If the input is an SSH certificate and `asCertificate` is false, the certificate
+// public key will be used as input for the fingerprint. If `asCertificate` is true,
+// the full contents of the certificate will be used in the fingerprint. If the input
+// is not an SSH certificate, but `asCertificate` is true, an error will be returned.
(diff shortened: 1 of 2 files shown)


hslatman:
Example cert:

```console
$ step ssh inspect example-cert.pub
example-cert.pub:
        Type: ecdsa-sha2-nistp256-cert-v01@openssh.com user certificate
        Public key: ECDSA-CERT SHA256:RvkDPGwl/G9d7LUFm1kmWhvOD9I/moPq4yxcb0STwr0
        Signing CA: ECDSA SHA256:AXEctpST7/1MfakrLrE+xrtF8Eixh6YsmqNaxiN6AFI (using ecdsa-sha2-nistp256)
        Key ID: "herman"
        Serial: 14376510....
```

Fingerprint before patching:

```console
$step ssh fingerprint example-cert.pub
256 SHA256:YuV7pyvW7Jp4iEwddOsMw+EdugrhKGqOKh6o+PP0xeg herman (ECDSA-CERT)
```

After patching:
```console
$ step ssh fingerprint example-cert.pub
256 SHA256:RvkDPGwl/G9d7LUFm1kmWhvOD9I/moPq4yxcb0STwr0 herman (ECDSA-CERT)
```

Output for SSH agent:
```console
$ ssh-add -l
256 SHA256:RvkDPGwl/G9d7LUFm1kmWhvOD9I/moPq4yxcb0STwr0 example (ECDSA)
256 SHA256:RvkDPGwl/G9d7LUFm1kmWhvOD9I/moPq4yxcb0STwr0 example (ECDSA-CERT)
```

maraino (commented):
I'm not sure about this.

For X.509 certificates we differentiate the fingerprint of a certificate than the fingerprint of the key. And getting the fingerprint of the key is as easy as inspecting the certificate.

SSH does shows the fingerprint of the key on both `ssh-add -l` or `ssh-keygen -l -f foo-cert.pub`.

hslatman:
My patched logic returns the same result as `ssh-add -l` when doing `step ssh list`. It confused @andysmallstep that `step ssh list` returned something different than `ssh-add -l` (while `ssh-add -L` and `step ssh list --raw` did/do have the same output), which prompted me to fix this. 

I could change it to only do this logic for `step ssh list`, but I think that'll be confusing in the end, because you would get different fingerprints for different use cases / commands.

The output of `step ssh fingerprint` _could_ be used when someone wants to configure a server with a trusted fingerprint. If it isn't the same as the one created by SSH, that may lead to confusion too.

I'm aware there's different logic for fingerprinting public keys vs. certificates for X.509, but I don't think that outweighs the fact that other tools seem to generate the fingerprint always for the public key from an SSH certificate.

maraino:
@hslatman, Are you going to create a new method to get the fingerprint for the certificate instead of the key as part of this PR?

hslatman:
Yes, will add it to this PR. Will be a new function, so that we don't need to change the signature of the current one, as you suggested. 

What do you think would be a good name for the flag (which will go in a PR for the CLI)? Something like `--full`, `--complete`, `--digest` or `--contents`? Or maybe just `--certificate`? The latter one would work better for `step ssh fingerprint`; less so for `step ssh list`. I like `--contents` and `--digest` best, I think. We could also add a flag `--alg` to change the algorithm used.

hslatman:
@maraino I added a new function for fingerprinting certificates: https://github.com/smallstep/crypto/pull/207/commits/a917ed948023ff426c83961b60a08d37d13befb6. I think it's good to be strict and to check the type, so that an error can be returned when one's trying to fingerprint a certificate, but operating on a key.

maraino (approved):

PR #208, opened 2023-03-29
Add support for asn1 methods in X509 templates

maraino:
### Description

This PR adds support for the following template functions:
* asn1enc, mustASN1Enc: given a string with or without a type, it encodes the ASN1 in base64.
* ans1Marshal, mustASN1Marshal: uses asn1.MarshalWithParams(), so it can support any type and options.
* asn1Seq, mustASN1Seq: given a list of base64(asn1) encoded strings, it builds a sequence of them, returning the base64 version of it.
* asn1Set, mustASN1Set: given a list of base64(asn1) encoded strings, it builds a set of them, returning the base64 version of it.

For example, you can build an extension like the following:
```go
{"id": "1.2.3.4", "value": {{ asn1Seq (asn1enc "YubiKey") (asn1enc "int:123456") | toJson }}}
```

### DIFF
--- a/x509util/extensions.go
+++ b/x509util/extensions.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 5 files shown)


x509util/extensions.go
```diff
@@ -439,28 +440,65 @@ func marshalOtherName(oid asn1.ObjectIdentifier, value interface{}) (asn1.RawVal
...
+			params = append(params, part)
+		// types that are parsed from the string.
+		// int and oid are not a type that can be set in a tag.
+		case "int", "oid":
+			p.Type = part
+		// types parsed from the string as a time
+		case "utc", "generalized":
+			p.Type = part
+			params = append(params, part)
+		// base64 encoded ans1 value
```
hslatman, line 466:
```suggestion
		// base64 encoded asn1 value
```

x509util/extensions.go
```diff
@@ -439,28 +440,65 @@ func marshalOtherName(oid asn1.ObjectIdentifier, value interface{}) (asn1.RawVal
...
+	return p
+}
+
+// marshalValue marshals the given value with the given params.
 //
 // The return value value can be any type depending on the OID ASN supports a great
 // number of formats, but Golang's ans1 package supports much fewer -- for now
 // support anything the golang asn1 marshaller supports.
 //
 // See https://www.openssl.org/docs/man1.0.2/man3/ASN1_generate_nconf.html
```
hslatman, line 485:
```suggestion
// marshalValue marshals the given value with the given params.
//
// The return value can be any type depending on the OID. ASN supports a great
// number of formats, but Golang's asn1 package supports much fewer -- for now
// support anything the Golang asn1 marshaller supports.
//
// See https://www.openssl.org/docs/man1.0.2/man3/ASN1_generate_nconf.html
```

x509util/options.go
```diff
@@ -81,3 +95,94 @@ func WithTemplateFile(path string, data TemplateData) Option {
...
+
+func asn1Set(b64enc ...string) string {
+	b64, err := mustASN1Set(b64enc...)
+	if err != nil {
+		return err.Error()
+	}
+	return b64
+}
+
+func mustASN1Encode(str string) (string, error) {
```
hslatman, line 131: The naming scheme is a bit surprising considering in that in other places where a `mustX` is used (usually in tests), no `error` is returned. I guess that's because when there's an error returned, it will be reported when the template is evaluated?

So in the case of `asn1Set`, an eventual error ends up as the value that's returned and thus possibly set in a template? It could lead to surprising results, I think. I wonder if we should just keep it strict, and only have it return an error in those cases (obviating the need for the `mustX`). Or maybe have a way for reporting it, but not via the template itself, but logged or otherwise returned?

maraino: Yes, the template interpreter fails when a function returns an error. Spring also uses `mustX` for this kind of function. But yes, it sounds good to me to fail always and remove the mustX methods.

x509util/options.go
```diff
@@ -36,7 +40,17 @@ func WithTemplate(text string, data TemplateData) Option {
 	return func(cr *x509.CertificateRequest, o *Options) error {
 		terr := new(TemplateError)
 		funcMap := templates.GetFuncMap(&terr.Message)
+		// asn1 methods
+		funcMap["asn1enc"] = asn1Encode
+		funcMap["asn1Marshal"] = asn1Marshal
+		funcMap["asn1Seq"] = asn1Sequence
+		funcMap["asn1Set"] = asn1Set
```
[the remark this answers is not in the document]
maraino: Fixed with 079c171: 
- `asn1Enc` only. 
- Remove `mustX` functions.
- Return an error in the rest.

x509util/extensions.go
maraino: Fixed with 079c171

x509util/extensions.go
maraino: Fixed with 079c171

x509util/options.go
maraino: Fixed with 079c171

hslatman (approved):

PR #223, opened 2023-04-21
Add support for key vault private DNS zone

maraino:
### Description

This commit adds support for using Key Vault keys using the private DNS zone instead of the public one.

Note that this PR might not be necessary, it looks that this is managed at the DNS server layer.


jdoss (approved):

maraino:
This is not necessary. Everything is done using CNAMEs at the DNS level.

PR #238, opened 2023-05-16
Add `CreateDecrypter` (RSA OAEP) support to Google Cloud KMS

hslatman:


kms/cloudkms/decrypter.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil, errors.New("cloudKMS AsymmetricDecrypt: request corrupted in-transit")
+	}
+	if int64(crc32c(response.Plaintext)) != response.PlaintextCrc32C.Value {
+		return nil, errors.New("cloudKMS AsymmetricDecrypt: response corrupted in-transit")
+	}
+
+	return response.Plaintext, nil
+}
+
+func crc32c(data []byte) uint32 {
```
maraino, line 115: Let's cast in this function instead of the other two places.

kms/cloudkms/decrypter.go
```diff
@@ -0,0 +1,120 @@
...
+	t := crc32.MakeTable(crc32.Castagnoli)
+	return crc32.Checksum(data, t)
+}
+
+var _ apiv1.Decrypter = (*CloudKMS)(nil)
```
maraino, line 120: What is this for?

kms/cloudkms/decrypter.go
```diff
@@ -0,0 +1,120 @@
...
+		switch ropts.Hash {
+		case crypto.SHA1, crypto.SHA256, crypto.SHA512:
+			break
+		default:
+			return nil, fmt.Errorf("cloudKMS does not support hash algorithm %q with RSA-OAEP", ropts.Hash)
+		}
+	}
+	if _, ok := opts.(*rsa.PKCS1v15DecryptOptions); ok {
+		return nil, errors.New("cloudKMS does not support PKCS #1 v1.5 decryption")
+	}
```
maraino, line 88: Let's add a typed switch here, similar to rsa.Decrypt(). The difference would be that if opts == nil the default algorithm would be RSA-OAEP instead of PKCS #1 v1.5. We should explicitly say that in the comment.

kms/cloudkms/decrypter.go
```diff
@@ -0,0 +1,120 @@
...
+// Google Cloud KMS decryption key.
+func NewDecrypter(client KeyManagementClient, decryptionKey string) (*Decrypter, error) {
+	// Make sure that the key exists.
+	decrypter := &Decrypter{
+		client:        client,
+		decryptionKey: decryptionKey,
+	}
+	if err := decrypter.preloadKey(decryptionKey); err != nil { // TODO(hs): (option for) lazy load instead?
+		return nil, err
+	}
```
maraino, line 44: Preloading the key makes sure that the key exists, and this is a fast operation.

maraino (changes requested):
Let's not support any type of DecrypterOpts

kms/cloudkms/decrypter.go
hslatman: Fixed in 1a83678b5bababd0f629cebd92ff7f49b8160ab8.

kms/cloudkms/decrypter.go
hslatman: It's called an interface guard. It makes it easier to verify that something implements a specific interface. Instead of relying on that at runtime or outside the package, a build will break if the interface is not implemented and your editor will also hint at it.

kms/cloudkms/decrypter.go
hslatman: I tried that, but the code doesn't look cleaner then. The reason for that is that there can be typed `nil`s, but also untyped `nil`s. Both have to be handled to be able to verify the input. 

The goal of the code is to show a more descriptive error message in case decryption fails if a user tries to use a parameter that we know will for sure result in an error. For example, when a label is used, the error is very generic: `Error: error decrypting input: cloudKMS AsymmetricDecrypt failed: rpc error: code = InvalidArgument desc = Decryption failed.`. Same is true if someone tries to use this with PKCS1 options. So we're not actually passing the options to GCP; we're merely checking for them, and return an error for things that GCP doesn't support.

kms/cloudkms/decrypter.go
hslatman: Yes, preloading absolutely makes sense. I'm not saying that we should remove that. Maybe "lazy load" isn't the right word for what I mean. Currently a failure to preload the key, for example due to a connection error, will result in the CA not starting, for example. It might be an option to not make this a hard fail immediately, but to try best effort to preload the key, and if it fails, to try it again when a key operation is required, for example. 

There's value in doing the preload and return an error, because it ensures the key is there, and there were no errors, so the decrypter (and signer) should be usable. But maybe we can decouple that operation, so that the user of the package can decide what semantics they want. For example, a failure to preload the public key could be handled when a call to `Public()` happens, so that the error can be logged instead of blocking CA initialization.

kms/cloudkms/decrypter.go
```diff
@@ -0,0 +1,120 @@
...
+		switch ropts.Hash {
+		case crypto.SHA1, crypto.SHA256, crypto.SHA512:
+			break
+		default:
+			return nil, fmt.Errorf("cloudKMS does not support hash algorithm %q with RSA-OAEP", ropts.Hash)
+		}
+	}
+	if _, ok := opts.(*rsa.PKCS1v15DecryptOptions); ok {
+		return nil, errors.New("cloudKMS does not support PKCS #1 v1.5 decryption")
+	}
```
maraino, line 89: You can do something like the following:

```go
func validateOAEPOptions(o *rsa.OAEPOptions) error {
	if len(o.Label) > 0 {
		return errors.New("cloudKMS does not support RSA-OAEP label")
	}
	switch o.Hash {
	case crypto.Hash(0), crypto.SHA1, crypto.SHA256, crypto.SHA512:
		return nil
	default:
		return fmt.Errorf("cloudKMS does not support hash algorithm %q with RSA-OAEP", ropts.Hash)
	}
}

func (d *Decrypter) Decrypt(rand io.Reader, ciphertext []byte, opts crypto.DecrypterOpts) ([]byte, error) {
	if opts == nil {
		opts = &rsa.OAEPOptions{}
	}

	switch o := opts.(type) {
	case *rsa.OAEPOptions:
		if err := validateOAEPOptions(o); err != nil {
			return nil, err
		}
	case *rsa.PKCS1v15DecryptOptions:
		return nil, errors.New("cloudKMS does not support PKCS #1 v1.5 decryption")
	default:
		return nil, errors.New("invalid options for Decrypt")
	}
	// ...
}
```

You can also do nothing in the `case *rsa.OAEPOptions` and do the validation after the switch.

The algorithm really depends on the key itself, so it makes sense to allow crypto.Hash(0) too.

hslatman: Changed it in c72666284c34ac267f1a5a83d0dcebbfece6354b. Note the additional `o == nil` case. It's required in a fairly specific usage pattern.

kms/cloudkms/decrypter.go
```diff
@@ -0,0 +1,137 @@
...
+// Also see https://cloud.google.com/kms/docs/algorithms#asymmetric_encryption_algorithms.
+func (d *Decrypter) Decrypt(rand io.Reader, ciphertext []byte, opts crypto.DecrypterOpts) ([]byte, error) {
+	if opts == nil {
+		opts = &rsa.OAEPOptions{}
+	}
+	switch o := opts.(type) {
+	case *rsa.OAEPOptions:
+		if o == nil { // var o *rsa.OAEPOptions; nothing to verify
+			break
+		}
```
maraino, line 98: you can do this in validateOAEPOptions.

maraino (approved):
lgtm

PR #239, opened 2023-05-18
Add `NewCertificateFromX509` function

hslatman:
The new function can create a `x509util.Certificate` from an x509.Certificate (used as a template). It also supports options, to provide a template, so that data can be set in the certificate.

Currently a template is always required. An error will be returned when no template is provided.

@maraino: I'll add tests later if this approach seems sensible to you. I don't think we can use generics to handle both `x509.Certificate` and `x509.CertificateRequest` options using only a single implementation, so I copied them over.


maraino:
I ❤️ the option of using generics.

hslatman:
Closing in favor of https://github.com/smallstep/crypto/pull/248.

PR #247, opened 2023-05-25
Make template functions generic with backwards compatibility

hslatman:
This is another approach to making the template functions generic, as done in #239, but now with backwards compatibility for users of the package.


hslatman:
Closing in favor of https://github.com/smallstep/crypto/pull/248.

PR #248, opened 2023-05-25
Add minimal `NewCertificateFromX509` implementation

hslatman:
Minimal implementation of `NewCertificateFromX509` to create an `x509.Certificate` without signing a CSR.

This PR replaces #239 and #247. We decided on keeping the logic simple, and to not add generic processing of `x509.Certificate` and `x509.CertificateRequest` for now.

The changes from the other PRs might still be of interest in the (near) future when more options are added for manipulating a `x509.Certificate` or `x509.CertificateRequest`, so we may reconsider the decision by then.


x509util/certificate.go
```diff
@@ -83,6 +83,58 @@ func NewCertificate(cr *x509.CertificateRequest, opts ...Option) (*Certificate,
...
+		ext, err := createCertificateSubjectAltNameExtension(cert, cert.Subject.IsEmpty())
+		if err != nil {
+			return nil, err
+		}
+		// Prepend extension to achieve a certificate as similar as possible to
+		// the one generated by the Go standard library.
+		cert.Extensions = append([]Extension{ext}, cert.Extensions...)
+	}
+
+	return &cert, nil
```
maraino, line 135: This is duplicated logic that can be combined into one internal function.

maraino (approved):
lgtm, but we should put the logic in one private method.

x509util/certificate.go
hslatman: Fixed in f1025e376b1005740f087c1103cf7605b20eb175

maraino (approved):
lgtm

PR #253, opened 2023-06-05
Add TPMKMS implementation

hslatman:
This PR adds a KMS implementation backed by a TPM. It uses our `tpm` package and its storage methods for its core operations. It uses URIs similar to the format used by PKCS11 to address / configure (attestation) keys.

The `TPMKMS` implementation supports attestation through configuration of the attestation CA to use when instantiating the KMS instance. This configuration is provided through an URI too. A default client implementation is provided that interacts with the Smallstep Attestation CA. A custom implementation can be provided through the `AttestationClient` interface.

As a bonus, this PR includes an `io.Reader` implementation backed by the `tpm` package to make a TPM generate random data.


kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+	akCert := akChain[0]
+	permanentIdentifier := ekKeyURL.String() // TODO(hs): should always match the valid value of the AK identity (for now)
+	resp := &apiv1.CreateAttestationResponse{
+		Certificate:         akCert,
+		CertificateChain:    akChain, // TODO(hs): should this include the leaf or not?
+		PublicKey:           akCert.PublicKey,
+		PermanentIdentifier: permanentIdentifier,
+	}
+
+	return resp, nil
```
maraino, line 450:
```suggestion
	return &apiv1.CreateAttestationResponse{
		Certificate:         akCert,
		CertificateChain:    akChain, // TODO(hs): should this include the leaf or not?
		PublicKey:           akCert.PublicKey,
		PermanentIdentifier: permanentIdentifier,
	}, nil
```

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+	if !hasValidIdentity(ak, ekKeyURL) {
+		return nil, fmt.Errorf("AK certificate (chain) not valid for EK %q", ekKeyURL)
+	}
+
+	// prepare the response to return
+	akCert := akChain[0]
+	permanentIdentifier := ekKeyURL.String() // TODO(hs): should always match the valid value of the AK identity (for now)
+	resp := &apiv1.CreateAttestationResponse{
+		Certificate:         akCert,
+		CertificateChain:    akChain, // TODO(hs): should this include the leaf or not?
```
maraino, line 445: Yubikey, which is the only one implementing this does not include the leaf.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+		return false
+	}
+	akCert := chain[0]
+
+	// TODO(hs): before continuing, add check if the cert is still valid?
+
+	// the Smallstep Attestation CA will issue AK certifiates that
+	// contain the EK public key ID encoded as an URN by default.
+	for _, u := range akCert.URIs {
+		if strings.EqualFold(ekURL.String(), u.String()) {
```
maraino, line 489: base64 is case-sensitive. Go's url.Parse always stores the schema in lower case, the only problem might be `sha256`, and I think it's ok to just use `==`

maraino (changes requested):

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
+//go:build !notpmkms
+// +build !notpmkms
```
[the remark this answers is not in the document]
hslatman: Done in 1686d1023a78d323f86c7ed848a5649617611e56.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed parsing %q: %w", req.Name, err)
+	}
+
+	v, ok := signatureAlgorithmMapping[req.SignatureAlgorithm]
+	if !ok {
+		return nil, fmt.Errorf("TPMKMS does not support signature algorithm %q", req.SignatureAlgorithm)
+	}
+
+	size := 2048
```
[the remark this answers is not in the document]
hslatman: I've seen most examples use 2048, which is also kind of a recommended minimum in the TPM specification. They discourage lower, but don't forbid it. For manufacturers following the spec, they might've implemented the bare minimum, which is why I set it to 2048. Also, AKs seem to be 2048 bits on Windows. It is possible to increase the size, but I think the template used to create the key will still set it to 2048 bits in one of the layers we build on.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+	permanentIdentifier   string
+}
+
+type algorithmAttributes struct {
+	Type  string
+	Curve int
+}
+
+var signatureAlgorithmMapping = map[apiv1.SignatureAlgorithm]algorithmAttributes{
+	apiv1.UnspecifiedSignAlgorithm: {"RSA", -1},
```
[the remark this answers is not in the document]
hslatman: Yes, that was my reasoning for picking RSA in this case.

kms/tpmkms/uri.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		return
+	}
+
+	if u, parseErr := uri.Parse(nameURI); parseErr == nil {
+		if u.Scheme != Scheme {
+			return o, fmt.Errorf("URI scheme %q is not supported", u.Scheme)
+		}
+	}
```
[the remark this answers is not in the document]
hslatman: It does do something: the first code path is followed when `err == nil`, indicating the scheme was correct. I wanted this to be a little stricter, and to only allow `tpmkms:`; no typos or other schemes allowed.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+		if err != nil {
+			if errors.Is(err, tpm.ErrExists) {
+				return nil, apiv1.AlreadyExistsError{Message: err.Error()}
+			}
+			return nil, fmt.Errorf("failed creating AK: %w", err)
+		}
+		createdAKURI := fmt.Sprintf("tpmkms:name=%s;ak=true", ak.Name())
+		return &apiv1.CreateKeyResponse{
+			Name: createdAKURI,
+		}, nil
```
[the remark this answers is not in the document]
hslatman: I've added the `PublicKey` in 1686d1023a78d323f86c7ed848a5649617611e56. If I remember correctly, AKs can sign data, but only data that originates from the TPM. I haven't implemented that in the `tpm` package yet.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+
+	createdKeyURI := fmt.Sprintf("tpmkms:name=%s", key.Name())
+	if properties.attestBy != "" {
+		createdKeyURI = fmt.Sprintf("%s;attest-by=%s", createdKeyURI, key.AttestedBy())
+	}
+
+	return &apiv1.CreateKeyResponse{
+		Name:       createdKeyURI,
+		PublicKey:  signer.Public(),
+		PrivateKey: priv,
```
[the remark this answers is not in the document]
hslatman: Fixed in 1686d1023a78d323f86c7ed848a5649617611e56.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+			SigningKey: createdKeyURI,
+		},
+	}, nil
+}
+
+// CreateSigner creates a signer using a key present in the TPM KMS.
+func (k *TPMKMS) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, error) {
+	if req.Signer != nil {
+		return req.Signer, nil
+	}
```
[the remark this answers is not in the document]
hslatman: Done in 1686d1023a78d323f86c7ed848a5649617611e56.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+	if cert == nil {
+		return nil, fmt.Errorf("failed getting certificate for %q: no certificate stored", properties.name)
+	}
+
+	return cert, nil
+}
+func (k *TPMKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
+	if req.Name == "" {
+		return errors.New("storeCertificateRequest 'name' cannot be empty")
+	}
```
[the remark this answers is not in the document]
hslatman: Fixed in 1686d1023a78d323f86c7ed848a5649617611e56.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+		return nil, errors.New("loadCertificateRequest 'name' cannot be empty")
+	}
+
+	properties, err := parseNameURI(req.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed parsing %q: %w", req.Name, err)
+	}
+
+	ctx := context.Background()
+	var cert *x509.Certificate // TODO(hs): support returning chain?
```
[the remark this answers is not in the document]
hslatman: I think we need it for TPMs. With the YubiKey, the attestation certificate is signed by the intermediate on the YubiKey, which can be retrieved from the YK again when the leaf has to be provided including the intermediate. With TPMs, we store the leaf + intermediate bundle originating from the attestation CA.

We could add a function for adding and loading chains. For YubiKeys, loading the chain would include the YK intermediate. Adding a chain on YK probably shouldn't work. This was an experimental interface, so I think a breaking change is OK in this case.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+		return nil, fmt.Errorf("failed getting EK public key ID: %w", err)
+	}
+	ekKeyURL := ekURL(ekKeyID)
+
+	// check if the derived EK URI fingerprint representation matches the provided
+	// permanent identifier value. The current implementation requires the EK URI to
+	// be used as the AK identity, so an error is returned if there's no match. This
+	// could be changed in the future, so that another attestation flow takes place,
+	// instead, for example.
+	if k.permanentIdentifier != "" && !strings.EqualFold(ekKeyURL.String(), k.permanentIdentifier) {
```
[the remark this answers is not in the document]
hslatman: Good one 😅  Fixed in 1686d1023a78d323f86c7ed848a5649617611e56.

kms/tpmkms/tpmkms.go
hslatman: In other places where we refer to a chain, we include the leaf as the first certificate. I think it's best to follow that? But if we do, `Certificate` is not required. In other places where attestation is done using certificates, I think there's usually a chain, which can be just the leaf. E.g. WebAuthn uses the `x5c` COSE claim for that. IMO defaulting to (just) the chain thus makes sense.

kms/tpmkms/tpmkms.go
hslatman: Fixed in 1686d1023a78d323f86c7ed848a5649617611e56. 

Currently no additional logic for `ek:sha256:`, but I could add that.

kms/tpmkms/tpmkms.go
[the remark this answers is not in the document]
hslatman: Would you be OK with enforcing AKs to be 2048 at all times? And for keys to use 3072 by default?

kms/tpmkms/tpmkms.go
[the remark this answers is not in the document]
maraino: The only marked as experimental is CreateAttestation. LoadCertificate is currently used in `step` for X5C tokens when you have the key and certificate in a KMS, currently we solve this using two flags `--x5c-cert` and `--x5c-chain`. And of course in `step-kms-plugin` that is the one really implementing the previous method.

For YubiKey and PKCS#11 we will need to keep supporting that case as they cannot store a chain. This is also implemented in CAPI, the current implementation is not using it, but apparently there's a [CertGetCertificateChain](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certgetcertificatechain) that we can use.

What is your proposal for keep supporting YubiKey, PKCS#11 and the ones supporting a chain? 

In StoreCertificate we can add a `CertificateChain []*x509.Certificate` what should YubiKey and PKCS#11 do?

Is adding [Load|Store]CertificateChain more simple? We can check if one kms implements one or the other and decide what to do.

kms/tpmkms/tpmkms.go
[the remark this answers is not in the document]
maraino: Is there a limitation? And yes it would be ok.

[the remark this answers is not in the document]
hslatman: In `go-attestation` the templates used for AKs are all 2048 bits. They can't be influenced directly, so AKs always need to be RSA and 2048, or there'll be a mismatch in what actually gets created. There's also some info on this in one of the documents in the TCG TPM specification. Fixed in the newer commits.

kms/tpmkms/tpmkms.go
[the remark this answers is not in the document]
hslatman: I've added a `CertificateChainManager` with `LoadCertificateChain` and `StoreCertificateChain`. Probably the simplest method and backwards compatible.

kms/tpmkms/tpmkms.go
hslatman: I've updated the `YubiKey` KMS to include the leaf as the first cert in the chain. This is not backwards compatible, but it's probably OK to do so.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+	apiv1.SHA512WithRSA:            {"RSA", -1},
+	apiv1.SHA256WithRSAPSS:         {"RSA", -1},
+	apiv1.SHA384WithRSAPSS:         {"RSA", -1},
+	apiv1.SHA512WithRSAPSS:         {"RSA", -1},
+	apiv1.ECDSAWithSHA256:          {"ECDSA", 256},
+	apiv1.ECDSAWithSHA384:          {"ECDSA", 384},
+	apiv1.ECDSAWithSHA512:          {"ECDSA", 521},
+}
+
+// New returns a new TPM KMS.
```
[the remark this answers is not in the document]
hslatman: Added in 514eb8a541579b1f9d0ce74fe30ea0f923463663.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,523 @@
...
+// CreateKey generates a new key in the TPM KMS and returns the public key.
+func (k *TPMKMS) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyResponse, error) {
+	switch {
+	case req.Name == "":
+		return nil, errors.New("createKeyRequest 'name' cannot be empty")
+	case req.Bits < 0:
+		return nil, errors.New("createKeyRequest 'bits' cannot be negative")
+	}
+
+	properties, err := parseNameURI(req.Name)
```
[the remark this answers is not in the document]
hslatman: Added in 514eb8a541579b1f9d0ce74fe30ea0f923463663.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,674 @@
...
+		return nil, fmt.Errorf("TPMKMS does not support signature algorithm %q", req.SignatureAlgorithm)
+	}
+
+	if properties.ak && v.Type == "ECDSA" {
+		return nil, errors.New("AKs must be RSA keys")
+	}
+
+	if properties.ak && req.Bits != defaultRSAAKSize { // 2048
+		return nil, fmt.Errorf("creating %d bit AKs is not supported; AKs must be RSA 2048 bits", req.Bits)
+	}
```
maraino, line 187: req.Bits == 0 should be valid if properties.ak is true.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,674 @@
...
+
+	signer, err := key.Signer(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed getting signer for key %q: %w", properties.name, err)
+	}
+
+	return signer.Public(), nil
+}
+
+func (k *TPMKMS) LoadCertificate(req *apiv1.LoadCertificateRequest) (*x509.Certificate, error) {
```
maraino, line 324: This method could call `LoadCertificateChain` and return the leaf.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	if len(chain) == 0 {
+		return nil, fmt.Errorf("failed getting certificate chain for %q: no certificate chain stored", properties.name)
+	}
+
+	return chain, nil
+}
+
+func (k *TPMKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
```
maraino, line 390: This method could call `StoreCertificateChain`

tpm/attestation/client.go
```diff
@@ -0,0 +1,282 @@
...
+
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			TLSClientConfig: &tls.Config{
+				RootCAs:            opts.rootCAs,
+				InsecureSkipVerify: opts.insecure, //nolint:gosec // intentional insecure if provided as option
+			},
+		},
```
maraino, line 87: We can clone the DefaultTransport that has good default and add the `TLSClientConfig`:

```go
transport := http.DefaultTransport.(*http.Transport).Clone()
transport.TLSClientConfig = &tls.Config{
    // ..
}
```

tpm/rand/rand.go
maraino, line 1: In the future we can consider adding another "Rander" interface to the kms.

maraino (changes requested):

kms/tpmkms/tpmkms.go
hslatman: Done in dc0a70717a587b95079a181a630c83126f150c2b.

kms/tpmkms/tpmkms.go
hslatman: Fixed in dc0a70717a587b95079a181a630c83126f150c2b 😓

tpm/rand/rand.go
hslatman: Not opposed to doing that, but I doubt that it should be part of the `kms` package or interface. Generating random data isn't something I would consider a KMS to be responsible for.

kms/apiv1/options.go
```diff
@@ -130,6 +143,10 @@ type Options struct {
 
 	// Profile to use in AmazonKMS.
 	Profile string `json:"profile,omitempty"`
+
+	// StorageDirectory is the path to a directory to
+	// store serialized TPM objects. Only used by the TPMKMS.
+	StorageDirectory string `json:"storageDirectory,omitempty"`
```
maraino, line 149: My preference is to remove all these extra options and have them configured through the URI. But I am ok to accepting this change.

tpm/rand/rand.go
maraino: I was suggesting it because you can get random bytes from PKCS#11, cloudkms and probably others.

kms/tpmkms/tpmkms.go
```diff
@@ -0,0 +1,658 @@
...
+// Attestation support in the TPMKMS is considered EXPERIMENTAL. It
+// is expected that there will be changes to the configuration that
+// be provided and the attestation flow.
+//
+// The TPMKMS implementation is backed by an instance of the TPM from
+// the `tpm` package. If the TPMKMS operations aren't sufficient for
+// your use case, use a tpm.TPM instance instead.
+func New(ctx context.Context, opts apiv1.Options) (kms *TPMKMS, err error) {
+	kms = &TPMKMS{}
+	storageDirectory := filepath.Join(step.Path(), "tpm") // store TPM objects in $STEPPATH/tpm by default
```
maraino, line 126: This doesn't take into account contexts, I suppose this is what we want. Right?

maraino (dismissed):
I've added a couple of comments, if you're ok with them, I am ok. It will also be good to send to codecov the coverage with the simulator.

kms/apiv1/options.go
hslatman: For programmatic configuration, this is a bit simpler to use. I figured a user of the package could be configuring this in their own program. Doing that through the URI is a bit more cumbersome than doing this. For the most common configurations, I think the property warranted. If we ever decide to remove it, users of the package will see the upgrade breaking their build, so I think it's OK for now.

hslatman:
> I've added a couple of comments, if you're ok with them, I am ok. It will also be good to send to codecov the coverage with the simulator.

I thought I already added coverage reporting through combining multiple tests runs, but maybe this typo resulted in it not working as expected: c2407d17eabc6309f7db535e5f1cb440f4d68c67.

maraino (dismissed):

maraino (approved):

PR #260, opened 2023-06-08
Add automatic support for p11-kit

### LINKED ISSUE #259: Use p11-kit-proxy.so as default provider where available
Ideally, users shouldn't need to specify a `module-path` in the URI and we could omit it from most of the examples. Most *nix systems have p11-kit which provides a system-wide configuration for PKCS#11 providers, and modern packaging of providers will include a p11-kit [module file](https://p11-glue.github.io/p11-glue/p11-kit/manual/pkcs11-conf.html) which makes it visible automatically.

So all I need to do to use a certificate/key from my Yubikey for a VPN connection, for example, is

```
$ sudo dnf install opensc
$ openconnect vpn.example.com --certificate 'pkcs11:manufacturer=piv_II;id=%01'
```

Having to specify a module-path is the pathological case. And even when it's needed, it would usually be *better* to fix the missing p11-kit config instead.

maraino:
### Description

This commit adds support to automatically use the p11-kit-proxy module on the PKCS#11 KMS if no other module has been specified.

Fixes #259

cc: @dwmw2

### DIFF
--- a/kms/pkcs11/pkcs11.go
+++ b/kms/pkcs11/pkcs11.go
@@ -54,39 +54,41 @@ type PKCS11 struct {
 
 // New returns a new PKCS11 KMS.
 func New(ctx context.Context, opts apiv1.Options) (*PKCS11, error) {
-	var config crypto11.Config
-	if opts.URI != "" {
-		u, err := uri.ParseWithScheme(Scheme, opts.URI)
-		if err != nil {
-			return nil, err
-		}
-
-		config.Pin = u.Pin()
-		config.TokenLabel = u.Get("token")
-		config.TokenSerial = u.Get("serial")
-		if v := u.Get("slot-id"); v != "" {
-			n, err := strconv.Atoi(v)
-			if err != nil {
-				return nil, errors.Wrap(err, "kms uri 'slot-id' is not valid")
-			}
-			config.SlotNumber = &n
-		}
-		// Get module or default to use p11-kit-proxy.so.
-		//
-		// pkcs11.New(module string) will use dlopen that will look for the
-		// given library in the appropriate paths, so there's no need to provide
-		// the full path.
-		if config.Path = u.Get("module-path"); config.Path == "" {
-			config.Path = defaultModule
-		}
+	if opts.URI == "" {
+		return nil, errors.New("kms uri is required")
 	}
+
+	var config crypto11.Config
+	u, err := uri.ParseWithScheme(Scheme, opts.URI)
+	if err != nil {
+		return nil, err
+	}
+
+	config.TokenLabel = u.Get("token")
+	config.TokenSerial = u.Get("serial")
+	if v := u.Get("slot-id"); v != "" {
+		n, err := strconv.Atoi(v)
+		if err != nil {
+			return nil, errors.Wrap(err, "kms uri 'slot-id' is not valid")
+		}
+		config.SlotNumber = &n
+	}
+
+	// Get module or default to use p11-kit-proxy.so.
+	//
+	// pkcs11.New(module string) will use dlopen that will look for the
+	// given library in the appropriate paths, so there's no need to provide
+	// the full path.
+	if config.Path = u.Get("module-path"); config.Path == "" {
+		config.Path = defaultModule
+	}
+
+	config.Pin = u.Pin()
 	if config.Pin == "" && opts.Pin != "" {
 		config.Pin = opts.Pin
 	}
 
 	switch {
-	case config.Path == "":
-		return nil, errors.New("kms uri 'module-path' are required")
 	case config.TokenLabel == "" && config.TokenSerial == "" && config.SlotNumber == nil:
 		return nil, errors.New("kms uri 'token', 'serial' or 'slot-id' are required")
 	case config.Pin == "":
@@ -114,8 +116,11 @@ func New(ctx context.Context, opts apiv1.Options) (*PKCS11, error) {
 var defaultModule = "p11-kit-proxy.so"
 
 func init() {
-	if runtime.GOOS == "darwin" {
+	switch runtime.GOOS {
+	case "darwin":
 		defaultModule = "p11-kit-proxy.dylib"
+	case "windows":
+		defaultModule = "p11-kit-proxy.dll"
 	}
 	apiv1.Register(apiv1.PKCS11, func(ctx context.Context, opts apiv1.Options) (apiv1.KeyManager, error) {
 		return New(ctx, opts)
--- a/kms/pkcs11/pkcs11.go
+++ b/kms/pkcs11/pkcs11.go
@@ -52,7 +52,30 @@ type PKCS11 struct {
 	closed sync.Once
 }
 
-// New returns a new PKCS11 KMS.
+// New returns a new PKCS#11 KMS. To initialize it, you need to provide a URI
+// with the following format:
+//
+//   - pkcs11:token=smallstep?pin-value=password
+//   - pkcs11:serial=1a2b3c4d5e6f?pin-source=/path/to/pin.txt
+//   - pkcs11:slot-id=5?pin-value=password
+//   - pkcs11:module-path=/path/to/module.so;token=smallstep?pin-value=password
+//
+// The scheme is "pkcs11"; "token", "serial", or "slot-id" defines the
+// cryptographic device to use, "module-path" is the path of the PKCS#11 module
+// to use, it will default to the proxy module of the p11-kit project if none is
+// specified (p11-kit-proxy.so), "pin-value" provides the user's PIN, and
+// "pin-source" defines a file that contains the PIN.
+//
+// A cryptographic key or object is identified by its "id" or "object"
+// attributes. The "id" is the key identifier for the object, it's a hexadecimal
+// string, and it will set the CKA_ID attribute of the object. The "object" is
+// the name of the object, and it will set the CKA_LABEL attribute. Only one
+// attribute is required to identify a key, but this package requires both to
+// create a new key. The complete URI for a key looks like this:
+//
+//   - pkcs11:token=smallstep;id=0a10;object=ec-key?pin-value=password
+//   - pkcs11:token=smallstep;id=0a10?pin-source=/path/to/pin.txt
+//   - pkcs11:token=smallstep;object=ec=key?pin-value=password
 func New(ctx context.Context, opts apiv1.Options) (*PKCS11, error) {
 	if opts.URI == "" {
 		return nil, errors.New("kms uri is required")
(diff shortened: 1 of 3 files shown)


kms/pkcs11/pkcs11_test.go
```diff
@@ -41,6 +41,18 @@ func TestNew(t *testing.T) {
...
 	}
 
+	var (
+		wantMissingModule    *PKCS11
+		wantErrMissingModule = true
+	)
+	if findP11KitProxy(context.Background()) != "" {
+		wantMissingModule = k
+		wantErrMissingModule = false
+	}
```
hslatman, line 51: This would be a more accurate test if the output of the shell command were mocked instead.

hslatman (commented):
The test could be improved, because right now the logic is (largely) dependent on the function that is being tested. I don't think it needs something very elaborate.

Other than that, looks good and interesting utility 😄

kms/pkcs11/pkcs11_test.go
maraino: I'll use `var findP11KitProxy = func(...) {...}`

dwmw2:
Reporting `p11-kit-proxy.so` instead of `.dylib` seems like a p11-kit bug. Is it reported and fixed upstream?

In some systems without development packages installed, pkg-config may be absent. Can we try just `p11-kit-proxy.(so|dylib)` with no path as a fallback?

As a follow-up, it would also be good to remove the explicit module-parh from most of the examples in the documentation, and make it clear that needing do so is a pathological case for a misconfigured system.

kms/pkcs11/pkcs11_test.go
maraino: Fixed with 237a370

hslatman (dismissed):

maraino:
@hslatman @dwmw2: In my last commit, 1e80f66, I removed the `pkg-config` code and rely on dlopen to find the right library in the default paths.

dwmw2:
Looks good to me; thanks for the prompt solution!

> People use this on containers where p11-kit is not available.

Well... I might argue that those come under the heading of 'misconfigured'. p11-kit is small and it *should* be present. And even when you only have one provider that you want to use, its packaging should still ship it with the p11-kit .module file and it shouldn't need to be manually specified.

But that's my [crusade](http://david.woodhou.se/draft-woodhouse-cert-best-practice.html), not yours :) And I'll begrudgingly accept the counter-argument if you have only one token then p11-kit, tiny though it may be, is overkill.

I'd still like to see the documentation at least mention that in a properly configured system, module-path *should* be unnecessary.

kms/pkcs11/pkcs11.go
```diff
@@ -70,6 +71,14 @@ func New(ctx context.Context, opts apiv1.Options) (*PKCS11, error) {
...
 			config.SlotNumber = &n
 		}
+		// Get module or default to use p11-kit-proxy.so.
+		//
+		// pkcs11.New(module string) will use dlopen that will look for the
+		// given library in the appropriate paths, so there's no need to provide
+		// the full path.
+		if config.Path = u.Get("module-path"); config.Path == "" {
+			config.Path = defaultModule
+		}
```
dwmw2, line 81: This is in the wrong place, isn't it?  It'll only happen if a URI is given, but there's an alternative code flow where it's all in the config? That should benefit from the default, too.

There's already a check for `config.Path == ""` a few lines further down, with an error return which is no longer telling the truth because it says that 'module-path is required'.

Also... what's wrong with `pkcs11:manufacturer=piv_II;id=%01` which I usually use as my examples in documentation? Why are only token/serial/slot-id accepted, and not model/manufacturer? 

(And why is this something you even need to do for yourself; surely "match token against the attributes in a URI" is a generic thing that ought to be somewhere much further down the stack than your code; in ThalesIgnite/crypto11 or miekg/pkcs11. Ideally you just want to call a function like the [matchSlots()](https://github.com/dwmw2/rolesanywhere-credential-helper/blob/ced1c51d62aeae24198ce489e02180c5ae1a025b/aws_signing_helper/pkcs11_signer.go#L66) I wrote here, and in your case as discussed, you want to just bail if `len(slots) != 1`).

dwmw2: > I'd still like to see the documentation at least mention that in a properly configured system, module-path should be unnecessary.

In particular, these two:
 * https://smallstep.com/docs/step-ca/configuration/#pkcs-11
 * https://github.com/smallstep/step-kms-plugin/blob/main/README.md
 
 .. especially the former. I don't want to live in a world where we have to tell users to explicitly pass a `module-path`.
  
 All those examples for how to use different HSMs differ mostly in the `module-path`, which shouldn't be needed anyway.
 
The packaging of the PKCS#11 provider module itself should come with a p11-kit .module file, the token should automatically appear in the output of things like `p11-kit list-tokens` or `p11tool --list-tokens` _(and thus also in GUIs  when you're selecting a client cert for 802.1x or VPN auth, etc.)_
 
 The common case is that user should just be able to select a token from the list, e.g. `pkcs11:manufacturer=piv_II`.

maraino: > This is in the wrong place, isn't it? It'll only happen if a URI is given, but there's an alternative code flow where it's all in the config? That should benefit from the default, too.

PKCS#11 cannot be configured without a URI, currently you will get the following error:
```
return nil, errors.New("kms uri 'module-path' are required")
```

We can make it more explicit and actually fail if URI == "".

> Also... what's wrong with pkcs11:manufacturer=piv_II;id=%01 which I usually use as my examples in documentation? Why are only token/serial/slot-id accepted, and not model/manufacturer?

Nothing, but https://github.com/ThalesIgnite/crypto11, which is the one we use on top of github.com/miekg/pkcs11 doesn't support it, see [this code](https://github.com/ThalesIgnite/crypto11/blob/a81014c7c41025fb5533c0c6b1b14bec016be695/crypto11.go#L285-L298).

Regarding the documentation, some of it is managed by a different team, and we haven't integrated this yet, so it will have to wait a bit. But I can change `step-kms-plugin` README and help once this is integrated.

kms/pkcs11/pkcs11.go
```diff
@@ -109,7 +109,14 @@ func New(ctx context.Context, opts apiv1.Options) (*PKCS11, error) {
 	}, nil
 }
 
+// defaultModule defines the defaultModule used, in this case is the
+// p11-kit-proxy provided by p11-kit.
+var defaultModule = "p11-kit-proxy.so"
+
 func init() {
+	if runtime.GOOS == "darwin" {
+		defaultModule = "p11-kit-proxy.dylib"
```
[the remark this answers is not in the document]
maraino: Good one.

>  Is there somewhere you can get the "correct" answer for the running system without having to hard-code it?

AFAIK there is not, and I cannot find anything in the stdlib to do it.

kms/pkcs11/pkcs11.go
maraino: I've fixed some of this in a23378d

dwmw2:
FWIW (although I don't think it matters any more) the `proxy_module` reported by p11-kit should be fixed with https://github.com/p11-glue/p11-kit/pull/516

kms/pkcs11/pkcs11.go
```diff
@@ -50,34 +52,66 @@ type PKCS11 struct {
...
+//
+//   - pkcs11:token=smallstep?pin-value=password
+//   - pkcs11:serial=1a2b3c4d5e6f?pin-source=/path/to/pin.txt
+//   - pkcs11:slot-id=5?pin-value=password
+//   - pkcs11:module-path=/path/to/module.so;token=smallstep?pin-value=password
+//
+// The scheme is "pkcs11"; "token", "serial", or "slot-id" defines the
+// cryptographic device to use, "module-path" is the path of the PKCS#11 module
+// to use, it will default to the proxy module of the p11-kit project if none is
+// specified (p11-kit-proxy.so), "pin-value" provides the user's PIN, and
```
hslatman, line 66:
```suggestion
// specified (p11-kit-proxy.so). "pin-value" provides the user's PIN, and
```

kms/pkcs11/pkcs11.go
```diff
@@ -50,34 +52,66 @@ type PKCS11 struct {
...
+// New returns a new PKCS#11 KMS. To initialize it, you need to provide a URI
+// with the following format:
+//
+//   - pkcs11:token=smallstep?pin-value=password
+//   - pkcs11:serial=1a2b3c4d5e6f?pin-source=/path/to/pin.txt
+//   - pkcs11:slot-id=5?pin-value=password
+//   - pkcs11:module-path=/path/to/module.so;token=smallstep?pin-value=password
+//
+// The scheme is "pkcs11"; "token", "serial", or "slot-id" defines the
+// cryptographic device to use, "module-path" is the path of the PKCS#11 module
```
hslatman, line 64:
```suggestion
// cryptographic device to use. "module-path" is the path of the PKCS#11 module
```

kms/pkcs11/pkcs11.go
```diff
@@ -50,34 +52,66 @@ type PKCS11 struct {
...
+// A cryptographic key or object is identified by its "id" or "object"
+// attributes. The "id" is the key identifier for the object, it's a hexadecimal
+// string, and it will set the CKA_ID attribute of the object. The "object" is
+// the name of the object, and it will set the CKA_LABEL attribute. Only one
+// attribute is required to identify a key, but this package requires both to
+// create a new key. The complete URI for a key looks like this:
+//
+//   - pkcs11:token=smallstep;id=0a10;object=ec-key?pin-value=password
+//   - pkcs11:token=smallstep;id=%0a%10?pin-source=/path/to/pin.txt
+//   - pkcs11:token=smallstep;object=ec=key?pin-value=password
```
hslatman, line 78:
```suggestion
//   - pkcs11:token=smallstep;id=0a10;object=ec-key?pin-value=password
//   - pkcs11:token=smallstep;id=%0a%10?pin-source=/path/to/pin.txt
//   - pkcs11:token=smallstep;object=ec-key?pin-value=password
```

kms/pkcs11/pkcs11.go
```diff
@@ -50,34 +52,66 @@ type PKCS11 struct {
...
+		return nil, err
+	}
+
+	config.TokenLabel = u.Get("token")
+	config.TokenSerial = u.Get("serial")
+	if v := u.Get("slot-id"); v != "" {
+		n, err := strconv.Atoi(v)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrap(err, "kms uri 'slot-id' is not valid")
```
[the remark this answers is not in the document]
maraino: The wrapped error contains the string.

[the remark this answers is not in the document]
hslatman: Fair enough

hslatman (approved):

PR #261, opened 2023-06-08
TPM improvements

hslatman:
Some fixes for KMS usage and additional functionality

### DIFF
--- a/kms/tpmkms/tpmkms.go
+++ b/kms/tpmkms/tpmkms.go
@@ -556,9 +556,9 @@ func (k *TPMKMS) CreateAttestation(req *apiv1.CreateAttestationRequest) (*apiv1.
 			return nil, fmt.Errorf("failed getting AK attestation parameters: %w", err)
 		}
 		return &apiv1.CreateAttestationResponse{
-			Certificate:         akChain[0],  // certificate for the AK
-			CertificateChain:    akChain,     // chain for the AK, including the leaf
-			PublicKey:           ak.Public(), // returns the public key of the attestation key
+			Certificate:         akChain[0], // certificate for the AK
+			CertificateChain:    akChain,    // chain for the AK, including the leaf
+			PublicKey:           akPub,      // returns the public key of the attestation key
 			PermanentIdentifier: permanentIdentifier,
 		}, nil
 	}
--- a/kms/tpmkms/tpmkms.go
+++ b/kms/tpmkms/tpmkms.go
@@ -345,6 +345,7 @@ func (k *TPMKMS) GetPublicKey(req *apiv1.GetPublicKeyRequest) (crypto.PublicKey,
 	return signer.Public(), nil
 }
 
+// LoadCertificate loads the certificate for the key identified by name from the TPMKMS.
 func (k *TPMKMS) LoadCertificate(req *apiv1.LoadCertificateRequest) (*x509.Certificate, error) {
 	if req.Name == "" {
 		return nil, errors.New("loadCertificateRequest 'name' cannot be empty")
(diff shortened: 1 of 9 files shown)


kms/tpmkms/tpmkms.go
```diff
@@ -483,6 +517,48 @@ func (k *TPMKMS) CreateAttestation(req *apiv1.CreateAttestationRequest) (*apiv1.
...
+		// TODO(hs): decide if we want/need to return these; their purpose is slightly
+		// different from the key certification parameters.
+		_, err = ak.AttestationParameters(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed getting AK attestation parameters: %w", err)
+		}
+		return &apiv1.CreateAttestationResponse{
+			Certificate:         akChain[0],  // certificate for the AK
+			CertificateChain:    akChain,     // chain for the AK, including the leaf
+			PublicKey:           ak.Public(), // returns the public key of the attestation key
```
maraino, line 557: You have `akPub` defined.

tpm/internal/key/key.go
```diff
@@ -108,6 +108,20 @@ type CreateConfig struct {
 	Size int
 }
 
+func (c *CreateConfig) Validate() error {
+	switch c.Algorithm {
+	case "RSA":
+		if c.Size > 2048 {
+			return fmt.Errorf("%d bits RSA keys are (currently) not supported; maximum is 2048", c.Size)
+		}
```
maraino, line 116: Is this strictly true? Is it because of go-attestation, or is it a limitation of the TPMs we have used?

tpm/key.go
```diff
@@ -185,6 +188,22 @@ func (t *TPM) CreateKey(ctx context.Context, name string, config CreateKeyConfig
...
 }
 
+type attestValidationWrapper attest.KeyConfig
+
+func (w attestValidationWrapper) Validate() error {
+	switch w.Algorithm {
+	case "RSA":
+		if w.Size > 2048 {
+			return fmt.Errorf("%d bits RSA keys are (currently) not supported; maximum is 2048", w.Size)
+		}
```
maraino, line 198: ditto

maraino (commented):
Should we not change DefaultRSASize back to 2048? It seems that more than that is not yet broadly supported.

tpm/internal/key/key.go
hslatman: I've added `go.step.sm/crypto` explicitly to the error message in faa54c3e06ca7e2dfca6fd7af81abfb5f1a0e585.

Could also prefix it, but `crypto` is less descriptive as to where it comes from by just reading the message.

The limitation is with the TPMs we've been testing with. At least that's my verdict right now. I see this error message as something temporary, until we add support for 1) checking the error resulting from the TPM and translating that to something more user friendly (ideally in a lower level, such as `go-tpm`) and/or 2) check the actual parameter values a specific TPM (model) allows before performing the operation (e.g. similar to the `tpm2_testparms` utility; and in a way that we can be sure that we're right about the capabilities of the TPM).

tpm/key.go
hslatman: Fixed in https://github.com/smallstep/crypto/commit/faa54c3e06ca7e2dfca6fd7af81abfb5f1a0e585.

hslatman:
> Should we not change DefaultRSASize back to 2048? It seems that more than that is not yet broadly supported.

Missed this one in the TPMKMS layer. Fixed in https://github.com/smallstep/crypto/pull/261/commits/5e479c17e0f2532d9a7336b89fb80042e7840d03.

maraino (approved):
Approved, but we can fail early on CreateKey

hslatman:
> Approved, but we can fail early on CreateKey

Let's revisit when we have a better approach to (in)valid TPM parameters

PR #267, opened 2023-06-19
Support for AK-only attestation through CreateAttestation

hslatman:
Before this PR AK attestation only worked for cases in which a (valid) AK certificate chain was already available, or when a new attestation was requested for a key attested by the AK. This PR adds support for obtaining an AK certificate for an AK if addressed directly (e.g. `tpmkms:name=my-ak;ak=true`) instead of addressing a key.

### DIFF
--- a/kms/tpmkms/tpmkms.go
+++ b/kms/tpmkms/tpmkms.go
@@ -534,9 +534,16 @@ func (k *TPMKMS) CreateAttestation(req *apiv1.CreateAttestationRequest) (*apiv1.
 	ekKeyURL := ekURL(ekKeyID)
 	permanentIdentifier := ekKeyURL.String()
 
+	// check if the derived EK URI fingerprint representation matches the provided
+	// permanent identifier value. The current implementation requires the EK URI to
+	// be used as the AK identity, so an error is returned if there's no match. This
+	// could be changed in the future, so that another attestation flow takes place,
+	// instead, for example.
+	if k.permanentIdentifier != "" && permanentIdentifier != k.permanentIdentifier {
+		return nil, fmt.Errorf("the provided permanent identifier %q does not match the EK URL %q", k.permanentIdentifier, permanentIdentifier)
+	}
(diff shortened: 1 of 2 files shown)


kms/tpmkms/tpmkms.go
```diff
@@ -534,58 +534,33 @@ func (k *TPMKMS) CreateAttestation(req *apiv1.CreateAttestationRequest) (*apiv1.
...
-		if err != nil {
-			return nil, fmt.Errorf("failed getting AK attestation parameters: %w", err)
-		}
-		return &apiv1.CreateAttestationResponse{
-			Certificate:         akChain[0], // certificate for the AK
-			CertificateChain:    akChain,    // chain for the AK, including the leaf
-			PublicKey:           akPub,      // returns the public key of the attestation key
-			PermanentIdentifier: permanentIdentifier,
-		}, nil
+		akName = key.AttestedBy()
```
maraino, line 556: Looks good, but I think it would be clear if you split the logic in two (if/else). They only share the call to `tpm.GetAK()`

```golang
if !provisioner.ak {
    ...
    // Do we need this? 
    // We already know the key was attested. 
    // And we're not using it in this flow.
    ak, err := k.tpm.GetAK(ctx, akName)
    if err != nil {
        return nil, err
    }
    ...
    return  &apiv1.CreateAttestationResponse{}, nil
}

// AK code
ak, err := k.tpm.GetAK(ctx, properties.name)
if err != nil {
    return  nil, err
}
...
return  &apiv1.CreateAttestationResponse{}, nil
```

maraino (changes requested):
Looks good, but I think we can clarify a little bit the flow, see my comment.

kms/tpmkms/tpmkms.go
maraino: Missed a code block.

maraino (approved):
lgtm

PR #273, opened 2023-06-21
Add UNIX socket TPM connection support

hslatman:
On Linux and macOS a connection can now be made to a TPM that's exposed on a UNIX socket. This shouldn't be used for production purposes, but can be useful for development and testing with a TPM simulator. The [tpmutil.EmulatorReadWriteCloser](https://pkg.go.dev/github.com/google/go-tpm/tpmutil#EmulatorReadWriteCloser) is used to create the connection with the simulated TPM. 

A UNIX socket TPM will only be used when the TPM is initialized with a device name that contains the path to a UNIX socket. If the TPM instance is initialized with a `tpm.CommandChannel` or `simulator.Simulator` directly, these will take precedence over the UNIX socket path.

Connecting to a TPM using a UNIX socket is not supported on Windows.

### DIFF
--- a/tpm/internal/socket/socket.go
+++ b/tpm/internal/socket/socket.go
@@ -1,11 +1,17 @@
 package socket
 
 import (
+	"errors"
 	"io"
 )
 
+var (
+	ErrNotAvailable = errors.New("socket not available")
+	ErrNotSupported = errors.New("connecting to a TPM using a UNIX socket is not supported on Windows")
+)
+
 func New(path string) (io.ReadWriteCloser, error) {
-	return new(path)
+	return newSocket(path)
 }
 
 type CommandChannelWithoutMeasurementLog struct {
--- a/tpm/internal/socket/socket.go
+++ b/tpm/internal/socket/socket.go
@@ -1,0 +1,17 @@
+package socket
+
+import (
+	"io"
+)
+
+func New(path string) (io.ReadWriteCloser, error) {
+	return new(path)
+}
+
+type CommandChannelWithoutMeasurementLog struct {
+	io.ReadWriteCloser
+}
+
+func (c *CommandChannelWithoutMeasurementLog) MeasurementLog() ([]byte, error) {
+	return nil, nil
+}
--- a/tpm/tpm.go
+++ b/tpm/tpm.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/smallstep/go-attestation/attest"
 
+	"go.step.sm/crypto/tpm/internal/close"
 	"go.step.sm/crypto/tpm/internal/open"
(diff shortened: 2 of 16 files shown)


tpm/internal/socket/socket.go
```diff
@@ -0,0 +1,23 @@
+package socket
+
+import (
+	"errors"
+	"io"
+)
+
+var (
+	ErrNotAvailable = errors.New("socket not available")
+	ErrNotSupported = errors.New("connecting to a TPM using a UNIX socket is not supported on Windows")
```
maraino, line 10: Sounds good.

There is limited support for UNIX sockets on [Windows](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/), but the proper solution are named pipes using https://github.com/microsoft/go-winio, for example, `step` connects to the agent using:

```golang
conn, err := winio.DialPipeContext(context.Background(), `\\.\\pipe\\openssh-ssh-agent`)
```

The same library can create a listener on a pipe using:
```golang
lis, err := winio.ListenPipe(`\\.\\pipe\\step-agent`, nil) 
```

If the simulator in windows can use those, they will not be difficult to implement in a future PR.

tpm/tpm.go
```diff
@@ -184,25 +273,29 @@ func (t *TPM) close(ctx context.Context) error {
...
 		}
 	}
 
 	// clean up the go-tpm rwc
 	if t.rwc != nil {
-		err := t.rwc.Close()
-		t.rwc = nil
-		if err != nil {
+		defer func() { t.rwc = nil }()
+		if err := closer.RWC(t.rwc); err != nil {
```
maraino, line 285: It might be ok, but it might be weird to set those to nil even if the close fails.

tpm/tpm.go
```diff
@@ -184,25 +273,29 @@ func (t *TPM) close(ctx context.Context) error {
...
 			return fmt.Errorf("failed closing rwc: %w", err)
 		}
 	}
 
 	return nil
 }
 
+func (t *TPM) Available() (err error) {
+	ctx := context.Background()
+	_, err = t.Info(ctx)
```
maraino, line 295: Nicer in one line:
```suggestion
	_, err = t.Info(context.Background())
```

maraino (dismissed):
LGTM, added a few comments.

tpm/tpm.go
hslatman: Yea, I kind of agree. But I think it's OK in this case, because no matter what happens, I want to have a new, clean `io.RWC` on the next `open`, and never start with an old one in a weird state.

maraino (approved):

PR #279, opened 2023-06-27
Upgrade `go-tpm` and adapt imports to use the legacy version

hslatman:
A new version of `go-tpm` [was released](https://github.com/google/go-tpm/releases/tag/v0.9.0) that changes the public API of the `tpm2` package. The `legacy/tpm2` contains the functions we used, so imports were adapted to use that, similar to what's done in other Google repos relying on `go-tpm`. 

The TPMDirect interface that was released with it should provide us a more pleasant and simpler approach for interfacing with the TPM, so I'll look into migrating our usage to that soon.


hslatman:
Looks like we'll have to wait till Go 1.21 is releases, so that we can drop Go 1.19 support.

hslatman:
After the Go version was upgraded to 1.20, the GH action now completes successfully.

maraino (approved):
lgtm

PR #294, opened 2023-07-24
Add support for early renewal for AK identity certificates

hslatman:
To support early renewal of AK certificates as part of a call to `CreateAttestation` in `TPMKMS`, the KMS should detect if an AK certificate is still within a specific percentage of it its original lifetime. This PR adds support for this check and introduces two new KMS URI options that influence the behavior:

* `disable-early-renewal=true`: disables the early renewal check and thus won't return an error when the certificate hasn't actually expired yet.
* `renewal-percentage=<int>`: sets the percentage of the lifetime after which the AK certificate will be considered to expire soon, and renewal will thus be tried.


maraino (approved):
lgtm

PR #298, opened 2023-07-26
Remove internal step package

maraino:
### Description

This commit removes the use of the step package and the use of STEPPATH as the base directory for relative paths in template files, as well as the location of the tpmkms storage directory.

This change will require changes in smallstep/cli and smallstep/certificates.


hslatman (approved):
Changes look good.

It probably shouldn't be merged before changes in certificates, CLI and plugins have been completed to ensure correct operation and to not accidentally get in with a library upgrade?

maraino:
@hslatman I've been making changes in other projects, and currently, the only problem I'm facing is that I need to specify "storage-directory" when using commands like:

```console
$ step path
/tmp/step
$ step kms key 'tpmkms:name=mariano;device=/var/run/step-tpmsimulator.sock;storage-directory=/tmp/step/tpm
-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEKYUDXQVk5lWPkN1TPy7lsAvQFp9O
ynHQW976bqcYCQB4bHNqr5j4dmWHYvIorHk5YDY3JbPAEb2jdTqSqos0iw==
-----END PUBLIC KEY-----
```

That flag was not necessary before.

PR #333, opened 2023-10-06
Add support for UPN SANs

maraino:
### Description

This PR adds a new type otherName SAN UPN with OID 1.3.6.1.4.1.311.20.2.3. This change also allows the use of the HardwareModuleName and DirectoryName SANs in the template. The previous version only allowed the use of those using code.

### DIFF
--- a/x509util/extensions.go
+++ b/x509util/extensions.go
@@ -66,6 +66,7 @@ const (
 	RegisteredIDType        = "registeredID"
 	PermanentIdentifierType = "permanentIdentifier"
 	HardwareModuleNameType  = "hardwareModuleName"
+	UPNType                 = "upn"
 )
 
 //nolint:deadcode // ignore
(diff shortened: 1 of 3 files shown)


x509util/extensions.go
```diff
@@ -66,6 +66,7 @@ const (
 	RegisteredIDType        = "registeredID"
 	PermanentIdentifierType = "permanentIdentifier"
 	HardwareModuleNameType  = "hardwareModuleName"
+	UPNType                 = "upn"
```
hslatman, line 69: Would propose to use `UserPrincipalNameType` internally instead of `UPNType`, but ain't blocking this from being used.

Besides that, maybe it should be `"userPrincipalName"` as the JSON property? That's more in line with the rest of the properties. Curious if it would be possible to use two names for the same, like an alias?

x509util/extensions.go
```diff
@@ -389,6 +412,22 @@ func (s SubjectAlternativeName) RawValue() (asn1.RawValue, error) {
 			IsCompound: true,
 			Bytes:      rdn,
 		}, nil
+	case UPNType:
+		if s.Value == "" {
+			return zero, errors.New("error parsing UserPrincipalName SAN: empty Value is not allowed")
```
hslatman, line 417: Minor, but I think more consistent with the error message below
```suggestion
			return zero, errors.New("error parsing UserPrincipalName SAN: empty value is not allowed")
```

x509util/extensions.go
```diff
@@ -389,6 +412,22 @@ func (s SubjectAlternativeName) RawValue() (asn1.RawValue, error) {
...
+		rawBytes, err := marshalExplicitValue(s.Value, "utf8")
+		if err != nil {
+			return zero, errors.Wrapf(err, "error marshaling ASN1 value %q", s.Value)
+		}
+		upnBytes, err := asn1.MarshalWithParams(otherName{
+			TypeID: oidUserPrincipalName,
+			Value:  asn1.RawValue{FullBytes: rawBytes},
+		}, "tag:0")
+		if err != nil {
+			return zero, errors.Wrap(err, "unable to Marshal UserPrincipalName SAN")
```
hslatman, line 428: In line with other (un)marshaling errors:
```suggestion
			return zero, errors.Wrap(err, "error marshaling UserPrincipalName SAN")
```

hslatman (commented):
This is great! Many MDMs support setting UPNs in their web interfaces.

x509util/extensions.go
maraino: Fixed with b724aa2

x509util/extensions.go
maraino: Fixed with b724aa2

x509util/extensions.go
maraino: Fixed with b724aa2

hslatman (approved):

PR #351, opened 2023-10-24
Add helpers to return the kms from an uri

maraino:
### Description

This commit introduces the helper `kms.TypeOf`, which returns the type of a URI. It also normalizes the type when necessary.

It also adds a commit to be able to validate JWKs that use an OpaqueSigner.

### DIFF
--- a/jose/validate.go
+++ b/jose/validate.go
@@ -162,6 +162,12 @@ func validateSigJWK(jwk *JSONWebKey) error {
 			return nil
 		}
 		errctx = "kty 'OKP' and crv 'Ed25519'"
+	case OpaqueSigner:
+		for _, alg := range k.Algs() {
+			if jwk.Algorithm == string(alg) {
+				return nil
+			}
+		}
 	}
 
 	return errors.Errorf("alg '%s' is not compatible with %s", jwk.Algorithm, errctx)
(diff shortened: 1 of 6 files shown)


kms/apiv1/registry.go
```diff
@@ -13,12 +13,12 @@ type KeyManagerNewFunc func(ctx context.Context, opts Options) (KeyManager, erro
...
 // Register adds to the registry a method to create a KeyManager of type t.
 func Register(t Type, fn KeyManagerNewFunc) {
-	registry.Store(t, fn)
+	registry.Store(t.normalize(), fn)
 }
 
 // LoadKeyManagerNewFunc returns the function initialize a KayManager.
 func LoadKeyManagerNewFunc(t Type) (KeyManagerNewFunc, bool) {
-	v, ok := registry.Load(t)
+	v, ok := registry.Load(t.normalize())
```
hslatman, line 21: Have a slight preference for doing the `t.normalize()` inside of `.Load` and `.Store` instead of the method call here, but not a blocker.

jose/validate.go
```diff
@@ -162,6 +162,12 @@ func validateSigJWK(jwk *JSONWebKey) error {
 			return nil
 		}
 		errctx = "kty 'OKP' and crv 'Ed25519'"
+	case OpaqueSigner:
+		for _, alg := range k.Algs() {
+			if jwk.Algorithm == string(alg) {
+				return nil
+			}
+		}
 	}
```
hslatman, line 171: I assume the default `errctx` value is sufficient for this case?

hslatman (approved):

jose/validate.go
maraino: Yes, that was the idea.

PR #358, opened 2023-11-03
TSS2 integration

maraino:
### Description

This PR adds the methods `ToTSS2()` to `tpm.AK` and `tpm.Key` to convert those keys to `tss2.TPMKey`. 

It also adds the `GetTSS2Signer()` method to the `tpm.TPM`, this will return a wrapped signer using the TSS2 key. This wrapped signer opens and closes the TPM on each sign.

And finally, it adds support for getting the `tss2.TPMKey` on `tpmkms.CreateKey`, and adds support to read TSS2 PEM files to `tpmkms.GetPublicKey` and `tpmkms.Createsigner`.


tpm/key.go
```diff
@@ -104,6 +105,15 @@ func (k *Key) MarshalJSON() ([]byte, error) {
 	return json.Marshal(o)
 }
 
+// ToTSS2 gets the public and private blobs and returns a [*tss2.TPMKey].
+func (k *Key) ToTSS2(ctx context.Context) (*tss2.TPMKey, error) {
```
hslatman, line 109: Somewhat related to my `tpm.TPM` remark: maybe a `ToTSS2Signer` also makes sense to add on `Key` and `AK`? I think it can get the reference to the `tpm.TPM` in one go that way?

tpm/signer.go
```diff
@@ -90,3 +91,41 @@ func (t *TPM) GetSigner(ctx context.Context, name string) (csigner crypto.Signer
...
+		return nil, fmt.Errorf("failed opening TPM: %w", err)
+	}
+	defer closeTPM(ctx, s.tpm, &err)
+	s.SetTPM(s.tpm.rwc)
+	signature, err = s.Signer.Sign(rand, digest, opts)
+	return
+}
+
+// GetTSS2Signer returns a crypto.Signer using the [tss2.TPMKey].
+func (t *TPM) GetTSS2Signer(ctx context.Context, key *tss2.TPMKey) (csigner crypto.Signer, err error) {
```
hslatman, line 114: I think this is not a great extension of the (public) API for the `tpm.TPM`. The reason is because it's now possible to provide a `tss2.TPMKey` that may or may not be (actually) backed by the `tpm.TPM` instance belonging to the TPM the TSS2 key was created for. Before this change, the `tpm.TPM` would be more of a facade to using the TPM, abstracting the underlying details such as key identifiers, storage format and attestation functionality. 

I understand that it's useful to get a reference to the TPMs `rwc`, but I think it's nicer to turn this dependency around? I think you'll still need it to be part of the `tpm` package, but it doesn't have to be a method on the `tpm.TPM`. Something like `Attach(key, tpm)` or `AttachTSS2(key, tpm)`, backed by roughly this method, but made private.

Another reason I think that's nicer is because we might still be able to implement a "full TSS2 format storage" that sits behind the `tpm.TPM` facade with all details abstracted.

tpm/tss2/signer.go
```diff
@@ -130,13 +136,30 @@ func CreateSigner(rw io.ReadWriter, key *TPMKey) (*Signer, error) {
...
+}
+
+// SetTPM allows to change the TPM channel. This operation is useful if the
+// channel set in [CreateSigner] is closed and opened again before calling [Signer.Sign].
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
+// release.
+func (s *Signer) SetTPM(rw io.ReadWriter) {
```
hslatman, line 151: Maybe call this `SetCommandChannel` instead? Although it's not entirely the same as `attest.CommandChannel`, but it sort of has that meaning in this context.

tpm/ak.go
```diff
@@ -457,6 +458,15 @@ func (ak *AK) HasValidPermanentIdentifier(permanentIdentifier string) bool {
 	return false
 }
 
+// ToTSS2 gets the public and private blobs and returns a [*tss2.TPMKey].
+func (ak *AK) ToTSS2(ctx context.Context) (*tss2.TPMKey, error) {
```
hslatman, line 462: See `Key`: add `ToTSS2Signer`?

kms/tpmkms/tpmkms.go
```diff
@@ -304,39 +326,76 @@ func (k *TPMKMS) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyRespons
 }
 
 // CreateSigner creates a signer using a key present in the TPM KMS.
+//
+// The `signingKey` in the [apiv1.CreateSignerRequest] can be used to specify
+// some key properties. These are as follows:
+//
+//   - name=<name>: specify the name to identify the key with
+//   - path=<file>: specify the TSS2 PEM file to use
```
hslatman, line 334: I'll think about this some more w.r.t. what it means for the `tpm.TPM` storage implementation, because there are files there too. I think adding `path` is OK as long as it's clear that those are for keys _outside_ of the TPM storage directory (usually). I would also like to keep the option open for a "native TSS2 storage".

tpm/signer.go
maraino: I don't get the name `Attach` why not just:
```go
func CreateTSS2Signer(ctx context.Context, t *TPM, key *tss2.TPMKey) (csigner crypto.Signer, err error)
```

tpm/key.go
maraino: Are you saying to add a `ToTSS2Signer` or change this name to `ToTSS2Signer`? 

We need a method to be able to transform a Key/AK to a `*tss2.TPMKey`, that type by itself is not a signer. And adding a `ToTSS2Signer()` doesn't really make a lot of sense, we already have `GetSigner()`:

```
func (t *TPM) GetSigner(ctx context.Context, name string) (csigner crypto.Signer, err error) 
```

kms/tpmkms/tpmkms.go
maraino: This is similar to the `path` used in `softkms`, where you can reference a file using `softkms:path=/dir/to/file.pem`. So yes, it's a file outside the TPM storage.

Yes, it also makes sense to create a native TSS2 storage, but we need to think a little bit more about it, because we will need to store certificates too. But once is done I imagine it will be similar to the current one and we will be using `name`.

tpm/tss2/signer.go
maraino: Fixed with 3f48274

tpm/signer.go
maraino: Changed to `CreateTSS2Signer` with 3f48274

kms/tpmkms/tpmkms.go
hslatman: The simplest approach might be to have a separate file (with the same or a derived name) with the cert next to the TSS2 PEM key. And possibly also a file with additional metadata, which could hold creation / attestation data.

tpm/signer.go
hslatman: It's just about naming things, and about making clear what it does.

I thought of `Attach` because the `tss2.Key` "attaches" to the `tpm.TPM`'s `rwc`, which makes it a complete `crypto.Signer`. The `tss2.Key` is (mostly) something external to the `tpm.TPM` (currently), so it somehow needs to be connected to it. It's also sufficiently different from the other methods, because it's a bit of a special case. Hence `Attach`. 

`CreateTSS2Signer` at the package level is better than what it was, but I don't like that it makes `TSS2` part of the top level public API surface. To me, at the moment, it's more of a utility method required because of how the `tpm` package is structured. I would also consider it experimental atm.

tpm/key.go
hslatman: `GetSigner` only works with AKs and Keys natively stored behind the TPM facade and identified by `name`, though. 

I was thinking about adding an additional `ToTSS2Signer` method, which would function similar to the (now) `CreateTSS2Signer`, but would be defined on the `Key` and `AK`, and then attach the reference to the `tpm.TPM` in that method immediately. Then you can get the signer without having to call `CreateTSS2Signer` explicitly.

hslatman (approved):

PR #363, opened 2023-11-08
Explicitly set destroyScheduledDuration to 24h

dopey:
- GCP is changing the default value from 24h to 30 days in a few months. This change makes sure that everything remains at the current value.


maraino (commented):
Not sure if this is the proper way to do it. We might want make this configurable.

dopey:
100% we want to make this configurable. However, this will at least keep things with the status quo for now. Otherwise it will switch to 30 days and then we'll be paying for destroyed keys even longer.

PR #387, opened 2023-12-12
Microsoft TPM Simulator TCTI (WIP)

hslatman:


hslatman:
go-tpm added support for the TCP TCTI: https://github.com/google/go-tpm/pull/387. Using that would require updating to the new `tpm2` package, and no longer depend on the `legacy` implementation.

PR #403, opened 2024-01-09
Implement yubikey connection cache

maraino:
This commit adds a cache of yubikey connections so the same card can be open more than once by the same application.

### DIFF
--- a/kms/yubikey/yubikey.go
+++ b/kms/yubikey/yubikey.go
@@ -32,6 +32,7 @@ var oidYubicoSerialNumber = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 41482, 3, 7}
 type YubiKey struct {
 	yk            pivKey
 	pin           string
+	card          string
 	managementKey [24]byte
 }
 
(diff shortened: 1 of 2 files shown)


kms/yubikey/yubikey.go
```diff
@@ -338,7 +359,11 @@ func (k *YubiKey) CreateAttestation(req *apiv1.CreateAttestationRequest) (*apiv1
 
 // Close releases the connection to the YubiKey.
 func (k *YubiKey) Close() error {
-	return errors.Wrap(k.yk.Close(), "error closing yubikey")
+	err := k.yk.Close()
+	if err == nil {
+		pivMap.Delete(k.card)
+	}
+	return errors.Wrap(err, "error closing yubikey")
```
hslatman, line 366: This surprised me a bit, but apparently `github.com/pkg/errors.Wrap` handles nil errs. Maybe slightly easier to understand if explicitly checking for it to be not nil?

maraino: Fixed with 4817292

hslatman (approved):

PR #413, opened 2024-01-23
Make `0x81000001` the default SRK handle when outputting TSS2

hslatman:


maraino (commented):
The change looks good, but some description in this PR of why this change is necessary.

If we run `openssl genpkey -provider tpm2 -algorithm RSA -out key.tss2`, the parent is `0x40000001`.

And in the last version of [TCG TSS 2.0 Overview and Common Structures Specification](https://trustedcomputinggroup.org/wp-content/uploads/TSS_Overview_Common_v1_r10_pub09232021.pdf) contradicts [Registry of Reserved TPM 2.0 Handles and Localities](https://trustedcomputinggroup.org/wp-content/uploads/Registry-of-Reserved-TPM-2.0-Handles-and-Localities-Version-1.2-Revision-1.00_pub.pdf), and still shows:
```c
#define TPM2_RH_OWNER        ((TPM2_RH) 0x40000001)
```

Note that the other one shows `TPM_RH_OWNER`, without the `2`.

hslatman:
> The change looks good, but some description in this PR of why this change is necessary.
> 
> If we run `openssl genpkey -provider tpm2 -algorithm RSA -out key.tss2`, the parent is `0x40000001`.
> 
> And in the last version of [TCG TSS 2.0 Overview and Common Structures Specification](https://trustedcomputinggroup.org/wp-content/uploads/TSS_Overview_Common_v1_r10_pub09232021.pdf) contradicts [Registry of Reserved TPM 2.0 Handles and Localities](https://trustedcomputinggroup.org/wp-content/uploads/Registry-of-Reserved-TPM-2.0-Handles-and-Localities-Version-1.2-Revision-1.00_pub.pdf), and still shows:
> 
> ```c
> #define TPM2_RH_OWNER        ((TPM2_RH) 0x40000001)
> ```
> 
> Note that the other one shows `TPM_RH_OWNER`, without the `2`.

Yes, that's correct. When using the TPM2 tools, by default and if you don't specify another parent, it would go directly under the owner handle. I'm not sure a different handle can be provided when using `openssl` with the `tpm2` provider, but you can provide a different handle when using e.g. `tpm2_create`. That's effectively what's used in `go-tpm` and `go-attestation`: a primary (the SRK) is created under `0x40000001` (with handle `0x81000001`), and then all AKs and keys are created under the SRK.

maraino (approved):
LGTM.

Perhaps makes sense to use 0x81000001 as the default.

hslatman:
As discussed: for external users of the `tss2` package, who may have obtained the pub/priv blobs from one of the `tpm2-tools`, the current default is to use the owner handle as the parent. We're keeping that in line in the `tss2` package, and only change it to `0x81000001` for TSS2 exports of our `AK` and `Key` types.

PR #419, opened 2024-02-06
Update attestation for a key

areed:
Currently a new key has to be created every time a new certificate is authorized with the acme device attestation protocol. This allows a new attestation to be generated with fresh qualifying data from the CA.

Linux

Windows

💔Thank you!


hslatman:
The test error is likely caused by not using the `ctx` and opening the TPM using the existing logic. There's a bit of plumbing to ensure the right command channel is set in all cases, incl. an in-process simulator for testing. See https://github.com/smallstep/crypto/blob/8684dc8bf25a0502f4c01cba205218c3cb09495a/tpm/key.go#L212C2-L215C30 for an example.

It's nice to see the new `tpm2` API being used, but I'm not sure if it's compatible with the existing plumbing as described above. If it is: great; otherwise would advise to use the `legacy` package.

PR #426, opened 2024-02-14
Do not enforce pin on pkcs11 kms

maraino:
This commit removes the enforcement of a pin when a pkcs11 kms is initialized. Some soft pkcs11 implementations, like NSS's soft token, have an empty pin by default.

### DIFF
--- a/kms/pkcs11/pkcs11.go
+++ b/kms/pkcs11/pkcs11.go
@@ -107,7 +107,7 @@ func New(_ context.Context, opts apiv1.Options) (*PKCS11, error) {
 	}
 
 	// We will allow empty pins as some modules might not have a pin by default.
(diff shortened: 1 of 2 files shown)


kms/pkcs11/pkcs11.go
```diff
@@ -106,6 +106,8 @@ func New(_ context.Context, opts apiv1.Options) (*PKCS11, error) {
 		config.Path = defaultModule
 	}
 
+	// We will allow empty pins as some modules might not have a pin by default.
+	// This is the case of softokn used to read NSS databases.
```
hslatman, line 110:
```suggestion
	// We will allow empty pins as some modules might not have a pin by default.
	// This is the case for softtoken, which is used to read NSS databases.
```

hslatman (approved):

PR #448, opened 2024-03-08
Add KMS for Apple keychain and Secure Enclave

maraino:
This commit adds the initial version of mackms, a kms for Apple keychain, and Secure Enclave. This version supports the KeyManager interface and an extra method for deleting keys in the Secure Enclave.


kms/mackms/mackms_test.go
```diff
@@ -0,0 +1,705 @@
...
+	"go.step.sm/crypto/randutil"
+)
+
+func createPrivateKeyOnly(t *testing.T, name string, signatureAlgorithm apiv1.SignatureAlgorithm) (*apiv1.CreateKeyResponse, error) {
+	t.Helper()
+
+	u, err := parseURI(name)
+	if err != nil {
+		return nil, err
+	}
```
hslatman, line 49: This instance and the others can be replaced with `require.NoError(t, err)` and not return the error.

kms/mackms/mackms_test.go
```diff
@@ -0,0 +1,705 @@
+//go:build darwin && cgo && !nomackms
```
hslatman, line 1: This probably needs the CI to be adapted to also run on macOS

kms/mackms/mackms_test.go
maraino: I wasn't sure if the defer functions were called, but I tested, and it does, `t.FailNow()` calls to `runtime.Goexit()`, which runs all deferred calls.

kms/mackms/mackms_test.go
maraino: This is something I considered doing but I decided to leave it for a future PR.

maraino:
@hslatman I've fixed most of your comments with 250ebe0, and change the MacKMS comment with d7a8c97

hslatman (approved):

dopey (approved):

PR #463, opened 2024-03-25
Add 'Serial' struct method returning serial number directly from Yubi…

joshdrake:
…Key.

This change provides a public struct method to obtain the serial number of a Yubikey directly from the the device; in the `piv` package this employs [the native PIV instruction](https://docs.yubico.com/yesdk/users-manual/application-piv/commands.html#get-the-serial-number) with a fallback for older models.

This enables client code using this library to always get the device serial number, even if the attestation key/certificate has been evicted/imported over.

💔Thank you!

### DIFF
--- a/kms/yubikey/yubikey.go
+++ b/kms/yubikey/yubikey.go
@@ -42,6 +42,7 @@ type pivKey interface {
 	GenerateKey(key [24]byte, slot piv.Slot, opts piv.Key) (crypto.PublicKey, error)
 	PrivateKey(slot piv.Slot, public crypto.PublicKey, auth piv.KeyAuth) (crypto.PrivateKey, error)
 	Attest(slot piv.Slot) (*x509.Certificate, error)
+	Serial() (uint32, error)
 	Close() error
 }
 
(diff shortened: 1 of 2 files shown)


kms/yubikey/yubikey.go
```diff
@@ -353,10 +354,20 @@ func (k *YubiKey) CreateAttestation(req *apiv1.CreateAttestationRequest) (*apiv1
...
 
+// Serial returns the serial number of the PIV card or and empty
+// string if retrieval fails
+func (k *YubiKey) Serial() string {
+	if serial, err := k.yk.Serial(); err == nil {
+		return strconv.FormatUint(uint64(serial), 10)
+	}
+
+	return ""
+}
```
maraino, line 369: You should return an error, wrapping it with something like:

```go
fmt.Errorf("error getting Yubikey's serial number: %w", err)
```

maraino (changes requested):

kms/yubikey/yubikey.go
joshdrake: ea7114e

maraino (approved):

PR #475, opened 2024-04-05
Support skipping certificate private key check on request

hslatman:
To support certificates that (usually) don't have a key present, `skip-find-certificate-key=true` can now be provided to an individual call for storing a certificate (chain). This can be helpful when storing an intermediate or root certificate in a specific store / store location.


kms/tpmkms/tpmkms.go
```diff
@@ -786,7 +790,7 @@ func (k *TPMKMS) storeCertificateChainToWindowsCertificateStore(req *apiv1.Store
 	}
 
 	if err := k.windowsCertificateManager.StoreCertificate(&apiv1.StoreCertificateRequest{
-		Name:        fmt.Sprintf("capi:sha1=%s;store-location=%s;store=%s;", fp, location, store),
+		Name:        fmt.Sprintf("capi:sha1=%s;store-location=%s;store=%s;skip-find-certificate-key=%s", fp, location, store, skipFindCertificatKey),
```
maraino, line 793: Are `location` and `store` properly encoded to work on a URI?

hslatman: For existing and valid values they are, when they're checked in the CAPI implementation. But I can change the implementation to `uv.Set`, or something similar, if preferred

hslatman: Fixed in 4d88201cc7a12b915cb355995878ae77be457e37

kms/capi/capi.go
```diff
@@ -714,6 +714,176 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
...
+			if bytes.Equal(x509Cert.SerialNumber.Bytes(), serialBytes) {
+				if err := windows.CertDeleteCertificateFromStore(certHandle); err != nil {
+					return fmt.Errorf("failed removing certificate: %w", err)
+				}
+
+				windows.CertFreeCertificateContext(certHandle)
+				return nil
+			}
+			prevCert = certHandle
+		}
```
maraino, line 881: We are not freeing all the certHandles in this loop.

maraino (changes requested):
We're not freeing all the certHandles, and I've also added another optional suggestion.

kms/uri/uri.go
```diff
@@ -150,13 +150,28 @@ func (u *URI) GetEncoded(key string) []byte {
...
+func (u *URI) GetHexEncoded(key string) []byte {
+	v := u.Get(key)
+	if v == "" {
+		return nil
+	}
+	if len(v)%2 == 0 {
+		if b, err := hex.DecodeString(strings.TrimPrefix(v, "0x")); err == nil {
+			return b
+		}
+	}
```
maraino, line 171: I think we should return an error in this case. You cannot know if there is an error unless you `Get` the key again.

kms/capi/capi.go
```diff
@@ -544,18 +544,15 @@ func (k *CAPIKMS) LoadCertificate(req *apiv1.LoadCertificateRequest) (*x509.Cert
...
 	switch {
-	case sha1Hash != "":
-		sha1Hash = strings.TrimPrefix(sha1Hash, "0x") // Support specifying the hash as 0x like with serial
-
-		sha1Bytes, err := hex.DecodeString(sha1Hash)
-		if err != nil {
-			return nil, fmt.Errorf("%s must be in hex format: %w", HashArg, err)
+	case len(sha1Hash) > 0:
+		if len(sha1Hash) != 20 {
+			return nil, fmt.Errorf("decoded %s has length %d; expected 20 bytes for SHA-1", HashArg, len(sha1Hash))
```
maraino, line 549: It's more customary to use bits for hashes, 160-bit in this case, but 20 bytes must be more clear.

kms/capi/capi.go
```diff
@@ -714,6 +711,172 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
...
+	}
+
+	var certStoreLocation uint32
+	switch storeLocation {
+	case "user":
+		certStoreLocation = certStoreCurrentUser
+	case "machine":
+		certStoreLocation = certStoreLocalMachine
+	default:
+		return fmt.Errorf("invalid cert store location %v", storeLocation)
```
maraino, line 745:
```suggestion
		return fmt.Errorf("invalid cert store location %q", storeLocation)
```

kms/capi/capi.go
```diff
@@ -714,6 +711,172 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
...
+	}
+
+	st, err := windows.CertOpenStore(
+		certStoreProvSystem,
+		0,
+		0,
+		certStoreLocation,
+		uintptr(unsafe.Pointer(wide(storeName))))
+	if err != nil {
+		return fmt.Errorf("CertOpenStore for the %v store %v returned: %w", storeLocation, storeName, err)
```
maraino, line 760:
```suggestion
		return fmt.Errorf("CertOpenStore for the %q store %q returned: %w", storeLocation, storeName, err)
```

kms/capi/capi.go
```diff
@@ -714,6 +711,172 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
...
+		if err := windows.CertDeleteCertificateFromStore(certHandle); err != nil {
+			return fmt.Errorf("failed removing certificate: %w", err)
+		}
+		return nil
+	case keyID != "":
+		keyID = strings.TrimPrefix(keyID, "0x") // Support specifying the hash as 0x like with serial
+
+		keyIDBytes, err := hex.DecodeString(keyID)
+		if err != nil {
+			return fmt.Errorf("%v must be in hex format: %w", KeyIDArg, err)
```
maraino, line 799: We can safely do %s:
```suggestion
			return fmt.Errorf("%s must be in hex format: %w", KeyIDArg, err)
```

maraino (changes requested):
GetHexEncoded should return an error if it is present but invalid. Nil if it is not present or empty.

kms/capi/capi.go
hslatman: When the value is specified, the value is in hex bytes, so I thought that would be clearer to understand in the error message.

maraino (approved):
lgtm

PR #483, opened 2024-04-12
Ignore non certificates in ReadCertificate and ReadCertificateBundle

dopey:


maraino (changes requested):
We can add variadic options to ReadCertificateBundle without sacrificing backward compatibility, but we should do this only if necessary, I don't think we do. 

By default, `ReadCertificateBundle` should get all `x509.Certificates` in a reasonable way. For example:

* Support empty spaces at the beginning, between certificates, and at the end (`step certificate inspect --bundle` already does that)
* Support for garbage date at the beginning, between certificates, and at the end (`step certificate inspect --bundle` already does that)
* Support for bundles in any order of certificates and any PEM formatted data like keys or ec parameters. (`step certificate inspect --bundle` FAILS with that)

The same can be applied to `ParseCertificate`, `ParseCertificateBundle`, `ParseCertificateRequest`, `ReadCertificate`, `ReadCertificateBundle`, `ReadCertificateRequest`, and `Parse` that will add support for others. 

Initially we can add support for any `*Certificate` and any `*CertificateBundle` methods.

pemutil/pem.go
```diff
@@ -280,34 +280,30 @@ func ParseCertificateRequest(pemData []byte) (*x509.CertificateRequest, error) {
 // ReadCertificate returns a *x509.Certificate from the given filename. It
 // supports certificates formats PEM and DER.
 func ReadCertificate(filename string, opts ...Options) (*x509.Certificate, error) {
-	b, err := utils.ReadFile(filename)
-	if err != nil {
+	// Populate options
+	ctx := newContext("PEM")
```
maraino, line 284: In this case it should be:
```suggestion
	ctx := newContext(filename)
```

pemutil/pem.go
```diff
@@ -280,34 +280,30 @@ func ParseCertificateRequest(pemData []byte) (*x509.CertificateRequest, error) {
...
-			return nil, errors.Errorf("error decoding PEM: file '%s' does not contain a certificate", filename)
-		}
+	bundle, err := ReadCertificateBundle(filename)
+	switch {
+	case err != nil:
+		return nil, err
+	case len(bundle) == 0:
+		return nil, errors.Errorf("file %s does not contain a valid PEM or DER formatted certificate", filename)
+	case len(bundle) > 1 && !ctx.firstBlock:
+		return nil, errors.Errorf("error decoding %s: contains more than one PEM encoded block", filename)
```
maraino, line 296: I guess we need this for backward compatibility. We can deprecate it in the future and use it by default.

pemutil/pem.go
```diff
@@ -280,34 +280,30 @@ func ParseCertificateRequest(pemData []byte) (*x509.CertificateRequest, error) {
...
-	// DER format (binary)
-	crt, err := x509.ParseCertificate(b)
-	return crt, errors.Wrapf(err, "error parsing %s", filename)
 }
 
 // ReadCertificateBundle returns a list of *x509.Certificate from the given
 // filename. It supports certificates formats PEM and DER. If a DER-formatted
 // file is given only one certificate will be returned.
+// This function will return an error if it encounters any non CERTIFICATE PEM
+// blocks in the file.
```
maraino, line 306: This is not true anymore.

maraino (changes requested):
Just a few notes. Other methods can also use this, but let's start with this.

maraino (approved):

PR #488, opened 2024-04-16
Allow extra content in CSR file for ReadCertificateRequest

dopey:


pemutil/pem.go
```diff
@@ -299,6 +300,16 @@ func ReadCertificate(filename string, opts ...Options) (*x509.Certificate, error
...
 
+type noValidPEMCertError struct {
+	filename string
+}
+
+func (e *noValidPEMCertError) Error() string {
+	return fmt.Sprintf("file %s does not contain a valid PEM formatted certificate", e.filename)
+}
+
+var ErrNoValidPEMCert *noValidPEMCertError
```
maraino, line 311: Something like can be more generic:
```suggestion
type invalidPEMFileError string

func (e invalidPEMFileError) Error() string {
	return fmt.Sprintf("file %s does not contain a valid PEM encoded data", string(e))
}

func (e invalidPEMFileError) Is(target error) bool {
	return target == ErrInvalidPEM
}

var ErrInvalidPEM = errors.New("invalid PEM encoded data")
```

You can also use a specific error for certificates if you want, but always with the `Is(error) bool` method, so you can do things like this:
```go
func foo() error {
	return invalidPEMFileError("file.crt")
}

func main() {
	if err := foo(); err != nil && errors.Is(err, ErrInvalidPEM {
		fmt.Fprintln(os.Stderr, "it doesn't look right")
	}
}
```

In some cases, you might also want to make the error public so the user can cast to it and get the `Filename`. But this can be added later. Later, we can generalize that to something more complex, for example things like:

```go
type InvalidPEMError struct {
	Type PEMType
	Filename string
	Reason string
}
```

maraino (changes requested):
At least we need to support the `errors.Is` method, but we should be adding your changes to other related methods in this package.

pemutil/pem.go
```diff
@@ -351,16 +362,19 @@ func ReadCertificateRequest(filename string) (*x509.CertificateRequest, error) {
...
+			if block == nil {
+				break
+			}
+			if !strings.HasSuffix(block.Type, "CERTIFICATE REQUEST") {
+				continue
+			}
+			csr, err := x509.ParseCertificateRequest(block.Bytes)
+			return csr, errors.Wrapf(err, "error parsing %s; CSR PEM block is invalid", filename)
 		}
+		return nil, errors.Errorf("file %s does not contain a valid PEM formatted CSR", filename)
```
maraino, line 377: This can be an custom error that can be compared with `ErrInvalidEM`

pemutil/pem.go
```diff
@@ -277,6 +278,52 @@ func ParseCertificateRequest(pemData []byte) (*x509.CertificateRequest, error) {
...
+func (pt PEMType) String() string {
+	switch pt {
+	case PEMTypeCertificate:
+		return "Certificate"
+	case PEMTypeCertificateRequest:
+		return "Certificate Request"
+	default:
+		return "Undefined"
+	}
+}
```
maraino, line 293: You're using this to rerturn an error, but it would be quite weird to use upper cases there, e.g, `file mariano.crt does not contain a valid PEM encoded Certificate"`

maraino (changes requested):

pemutil/pem.go
```diff
@@ -277,6 +278,56 @@ func ParseCertificateRequest(pemData []byte) (*x509.CertificateRequest, error) {
...
+	switch {
+	case e.Message != "":
+		return e.Message
+	case e.Err != nil:
+		return e.Err.Error()
+	default:
+		if e.Type == PEMTypeUndefined {
+			return fmt.Sprintf("file %s does not contain valid PEM encoded data", e.File)
+		}
+		return fmt.Sprintf("file %s does not contain a valid PEM encoded %s", e.File, e.Type)
```
maraino, line 323: It works for now, but the file might not be present if we add this error to methods like `ParseCertificate` where the arguments are just the PEM bytes.

pemutil/pem.go
```diff
@@ -351,16 +402,20 @@ func ReadCertificateRequest(filename string) (*x509.CertificateRequest, error) {
...
+			block, b = pem.Decode(b)
+			if block == nil {
+				break
+			}
+			if !strings.HasSuffix(block.Type, "CERTIFICATE REQUEST") {
+				continue
+			}
+			csr, err := x509.ParseCertificateRequest(block.Bytes)
+			return csr, &InvalidPEMError{File: filename, Type: PEMTypeCertificateRequest,
+				Err: fmt.Errorf("error parsing %s: CSR PEM block is invalid: %w", filename, err)}
```
maraino, line 416: It would make more sense to set that as the message and set `Err: err`

pemutil/pem.go
```diff
@@ -351,16 +407,22 @@ func ReadCertificateRequest(filename string) (*x509.CertificateRequest, error) {
...
+			block, b = pem.Decode(b)
+			if block == nil {
+				break
+			}
+			if !strings.HasSuffix(block.Type, "CERTIFICATE REQUEST") {
+				continue
+			}
+			csr, err := x509.ParseCertificateRequest(block.Bytes)
+			return csr, &InvalidPEMError{File: filename, Type: PEMTypeCertificateRequest,
+				Message: fmt.Errorf("error parsing %s: CSR PEM block is invalid: %w", filename, err).Error(),
```
maraino, line 421:
```suggestion
				Message: fmt.Sprintf("error parsing %s: CSR PEM block is invalid: %v", filename, err),
```

maraino (changes requested):
Just an small change

pemutil/pem.go
```diff
@@ -277,6 +278,60 @@ func ParseCertificateRequest(pemData []byte) (*x509.CertificateRequest, error) {
...
+	Message string
+	Err     error
+}
+
+func (e *InvalidPEMError) Error() string {
+	switch {
+	case e.Message != "":
+		return e.Message
+	case e.Err != nil:
+		return fmt.Errorf("error decoding PEM data: %w", e.Err).Error()
```
maraino, line 318:
```suggestion
		return fmt.Sprintf("error decoding PEM data: %v", e.Err)
```

maraino (changes requested):

maraino (approved):
lgtm

PR #490, opened 2024-04-24
Support reading from STDIN in `ReadFile`

hslatman:
Package `utfbom` is a copy of https://github.com/smallstep/cli/tree/master/utils/internal/utfbom. The `ReadFile` method is now more like https://github.com/smallstep/cli/blob/master/utils/read.go#L88.


merlinz01:
Am I right that this will allow me to do `step ca certificate [args] --provisioner-password-file -` and send the password to stdin? If so I will be glad to see this merged.
It will make it very nice for automating that command using Salt Stack. I was trying to do it by specifying `--provisioner-password-file /dev/stdin`, but that makes issues because of `/dev/stdin` permissions when Salt runs it as a different user.

hslatman:
@merlinz01 yes, there's several function calls in between, but I believe it will indeed eventually end up at `ReadFile` through `ReadPasswordFromFile`.

dopey (approved):

PR #503, opened 2024-05-14
Require ClientAuth when verifying an X5cInsecure certificate

maraino:
The X5cInsecure certificate is used by step-ca to renew certificates without using mTLS, usually expired certificates. `Certificate.Verify` defaults to require ServerAuth if no KeyUsages is set as an option. Due to how these tokens are used, it makes more sense to require only ClientAuth.

Note that a few lines after this check is also made:
```go
if leaf.KeyUsage&x509.KeyUsageDigitalSignature == 0 {
	return nil, nil, errors.New("certificate used to sign x5cInsecure token cannot be used for digital signature")
}
```

Related to smallstep/certificates#1843

### DIFF
--- a/jose/parse.go
+++ b/jose/parse.go
@@ -267,6 +267,9 @@ func ParseX5cInsecure(tok string, roots []*x509.Certificate) (*JSONWebToken, [][
 		Intermediates: interPool,
 		// A hack so we skip validity period validation.
 		CurrentTime: leaf.NotAfter.Add(-1 * time.Minute),
+		KeyUsages: []x509.ExtKeyUsage{
+			x509.ExtKeyUsageClientAuth,
+		},
 	})
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "error verifying x5cInsecure certificate chain")


jose/parse.go
```diff
@@ -267,6 +267,9 @@ func ParseX5cInsecure(tok string, roots []*x509.Certificate) (*JSONWebToken, [][
 		Intermediates: interPool,
 		// A hack so we skip validity period validation.
 		CurrentTime: leaf.NotAfter.Add(-1 * time.Minute),
+		KeyUsages: []x509.ExtKeyUsage{
+			x509.ExtKeyUsageClientAuth,
+		},
```
hslatman, line 272: Should `ExtKeyUsageServerAuth` be provided too, for backwards compatibility? I agree that it makes more sense to require just client auth, but there's a chance there's certs out there being used that don't have it set.

maraino: I don't think so. This method is used as a replacement for mTLS when this cannot be performed, and the requirement for a client to do mTLS is to have `x509.ExtKeyUsageClientAuth`. See
https://github.com/golang/go/blob/1f6a983baf4b9a636e9e4bbd827fcb4d6ef4ebe0/src/crypto/tls/handshake_server.go#L892-L897

hslatman: I'm fine with merging it like this, but it still can be a breaking change in practice. Likely a fairly small chance, but still.

hslatman (approved):

PR #515, opened 2024-05-30
Add support for tapping into TPM communication

hslatman:


tpm/debug/pcapng.go
```diff
@@ -0,0 +1,188 @@
...
+var decodeOptions = gopacket.DecodeOptions{
+	Lazy:   true,
+	NoCopy: true,
+}
+
+func write(w *pcapgo.NgWriter, data []byte, in bool, inSeq, outSeq uint32) error {
+	var tcpLayer = &layers.TCP{Window: 16}
+	if in {
+		tcpLayer.SrcPort = layers.TCPPort(2321)
+		tcpLayer.DstPort = layers.TCPPort(50001)
```
maraino, line 150: Didn't look too much into this PR, but we probably want to move these ports customizable.

hslatman: We could, but that wouldn't change much functionally. The PCAP is used just as a container in this case; we're not actually capturing traffic from/to those ports. The bytes read from and written to the TPM device are "massaged into network packets", so to say.

PR #534, opened 2024-06-28
Add support validities in templates

maraino:
### Description

This commit allows setting the validity bounds on X.509 and SSH certificates using templates. 

It also adds the template method `toTime`, which makes it easier to format a timestamp to RFC3339. This format is compatible with Go's `time.Time` when `json.Unmarshal` is used. There are similar methods to do this in the sprig library, but they require you to specify the format string. For example, these methods will return a valid string that can be unmarshaled to `time.Time`:

```go
{{ date "2006-01-02T15:04:05Z07:00" .Token.nbf | toJson }}
{{ dateInZone "2006-01-02T15:04:05Z07:00" .Token.nbf "UTC" | toJson }}
{{ toTime .Token.nbf | toJson }}
```

### DIFF
--- a/internal/templates/funcmap_test.go
+++ b/internal/templates/funcmap_test.go
@@ -3,6 +3,11 @@ package templates
 import (
 	"errors"
 	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.step.sm/crypto/jose"
 )
 
 func Test_GetFuncMap_fail(t *testing.T) {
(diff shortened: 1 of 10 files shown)


internal/templates/funcmap_test.go
```diff
@@ -20,3 +25,51 @@ func Test_GetFuncMap_fail(t *testing.T) {
...
+
+	t.Run("default", func(t *testing.T) {
+		var failMesage string
+		fns := GetFuncMap(&failMesage)
+		fn := fns["toTime"].(func(any) string)
+		want := time.Now()
+		got, err := time.Parse(time.RFC3339, fn(nil))
+		require.NoError(t, err)
+		assert.WithinDuration(t, want, got, time.Second)
+		assert.Equal(t, got.Location(), time.UTC)
```
hslatman, line 73:
```suggestion
		assert.Equal(t, time.UTC, got.Location())
```

x509util/certificate.go
```diff
@@ -165,6 +168,10 @@ func (c *Certificate) GetCertificate() *x509.Certificate {
 		e.Set(cert)
 	}
 
+	// Validity bpunds.
```
hslatman, line 171:
```suggestion
	// Validity bounds.
```

hslatman (commented):
Minor doc issues

hslatman (dismissed):

hslatman:
That lint issue is my bad with the GH editor

hslatman (approved):

PR #552, opened 2024-07-18
Add `SearchKeys` functionality to MacKMS

hslatman:
In follow up PRs I'd like to add:

- Support in PKCS11
- Support in TPMKMS
- Support in CAPI?
- Support for searching certificates?

### DIFF
--- a/kms/mackms/mackms.go
+++ b/kms/mackms/mackms.go
@@ -555,7 +555,7 @@ func (*MacKMS) DeleteCertificate(req *apiv1.DeleteCertificateRequest) error {
 	return nil
 }
 
-func (*MacKMS) SearchKeys(req *apiv1.SearchKeysRequest) (*apiv1.SearchKeysResponse, error) {
+func (k *MacKMS) SearchKeys(req *apiv1.SearchKeysRequest) (*apiv1.SearchKeysResponse, error) {
 	if req.Query == "" {
 		return nil, fmt.Errorf("searchKeysRequest 'query' cannot be empty")
 	}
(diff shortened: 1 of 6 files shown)


kms/mackms/mackms.go
```diff
@@ -555,6 +555,73 @@ func (*MacKMS) DeleteCertificate(req *apiv1.DeleteCertificateRequest) error {
...
+
+	keys, err := getPrivateKeys(u)
+	if err != nil {
+		return nil, fmt.Errorf("failed getting keys: %w", err)
+	}
+
+	results := make([]apiv1.SearchKeyResult, len(keys))
+	for i, key := range keys {
+		d := cf.NewDictionaryRef(cf.TypeRef(key.TypeRef()))
+		defer d.Release()
```
maraino, line 588: Does `d.Release()` release its contents?

When building the query dictionary, we release both contents and the dictionary.

kms/mackms/mackms.go
```diff
@@ -555,6 +555,73 @@ func (*MacKMS) DeleteCertificate(req *apiv1.DeleteCertificateRequest) error {
 	return nil
 }
 
+// SearchKeys searches for keys according to the query URI in the request.
+//
+//   - "" will return all keys managed by the KMS (using the default tag)
+//   - "mackms:" will return all keys managed by the KMS  (using the default tag)
+//   - "mackms:label=my-label" will return all keys using label "my-label" (and the default tag)
+//   - "mackms:hash=the-hash" will return all keys having hash "hash" (and the default tag; generally one result)
+//   - "mackms:tag=my-tag" will search for all keys with "my-tag"
```
maraino, line 564: Would it be possible to search all the keys in the secure enclave that matches the tag/label/hash? 

This `mackms:se=true` or this `mackms:tag=my-tag;se=true` might be useful.

kms/mackms/mackms.go
```diff
@@ -625,6 +692,70 @@ func getPrivateKey(u *keyAttributes) (*security.SecKeyRef, error) {
...
+	}
+
+	array := cf.NewArrayRef(result)
+	defer array.Release()
+
+	keys := make([]*security.SecKeychainItemRef, array.Len())
+	for i := 0; i < array.Len(); i++ {
+		item := array.Get(i)
+		key := security.NewSecKeychainItemRef(item)
+		key.Retain() // retain the key, so that it's not released early
```
maraino, line 752: I suppose this gets released early because of the `array.Release()`, right?

maraino (commented):
Looks ok, I've added a few questions and a comment to see if it is possible to search keys in the secure enclave. For example, all with the default tag would be `mackms:se=true`

kms/mackms/mackms.go
hslatman: Not retaining it here will result in an error on getting the first property from the dictionary (note that it fails on `GetSecAttrTokenID`, after some reordering):

```
runtime.cgocall(0x100a5e870, 0x1400012da28)
	/Users/herman/sdk/go1.21.0/src/runtime/cgocall.go:157 +0x44 fp=0x1400012d9f0 sp=0x1400012d9b0 pc=0x100709fc4
go.step.sm/crypto/internal/darwin/security._Cfunc_CFDictionaryGetValue(0x12821a9e0, 0x1e5d71890)
	_cgo_gotypes.go:363 +0x34 fp=0x1400012da20 sp=0x1400012d9f0 pc=0x10096dae4
go.step.sm/crypto/internal/darwin/security.GetSecAttrTokenID.func1(0xffffffffffffffff?)
```

kms/mackms/mackms.go
hslatman: I'll add `defer key.Release()` here

hslatman: Fixed in 606fcd9b12dd745665f5469b900de70568d31610.

kms/mackms/mackms.go
hslatman: I've added support for this in 606fcd9b12dd745665f5469b900de70568d31610. Right now I'm doing the check in the Go code. It might be possible to set the attribute in the query, but it's not so simple to test the actual Secure Enclave operations, so that's why I chose this option.

When `se` is set, its value will be used to filter the results that are returned. If it's not set in the search query, it will be ignored, and both types of keys will be returned.

maraino (approved):
LGTM.

In the switch cases, we could remove the se=false in the name URI, which would be the default for individual keys.

PR #560, opened 2024-07-31
tpm/storage: Blackhole rework

azazeal:
This PR reworks the `Blackhole` TPM store so that it returns a sentinel error for its mutations and always `ErrNotFound` for its queries.


azazeal:
Closed in favor of #561

PR #561, opened 2024-07-31
Return a sentinel error when no storage is configured when required

hslatman:
The goal of the blackhole storage is to be able to use an instance of `tpm.TPM` for operations that don't require an (metadata) storage. In some usages it could result in an error or even a panic, because it's not clear enough that some operations require an actual storage implementation for the operations to succeed. 

This commit changes the behavior of the blackhole storage. It now returns a sentinel error for operations that require a storage to be successful. The sentinel error is handled in the `tpm` package wherever a storage operation takes place.

### DIFF
--- a/tpm/tpm_test.go
+++ b/tpm/tpm_test.go
@@ -7,6 +7,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+
+	"go.step.sm/crypto/tpm/storage"
 )
 
 type closeSimulator struct {
@@ -73,4 +75,10 @@ func Test_close(t *testing.T) {
 	require.Nil(t, tpm.simulator)
 	closeTPM(context.Background(), newCloseErrorTPM(t), &closeErr)
 	require.EqualError(t, closeErr, "failed closing attest.TPM: closeErr") // attest.TPM is backed by the closeSimulator
+}
+
+func TestTPMNoStorageConfiguredError(t *testing.T) {
+	err := ErrNoStorageConfigured
+	require.ErrorIs(t, err, ErrNoStorageConfigured)
+	require.ErrorIs(t, err, storage.ErrNoStorageConfigured)
 }
--- a/tpm/tpm_test.go
+++ b/tpm/tpm_test.go
@@ -79,6 +79,5 @@ func Test_close(t *testing.T) {
 
 func TestTPMNoStorageConfiguredError(t *testing.T) {
 	err := ErrNoStorageConfigured
-	require.ErrorIs(t, err, ErrNoStorageConfigured)
 	require.ErrorIs(t, err, storage.ErrNoStorageConfigured)
 }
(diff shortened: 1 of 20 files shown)


hslatman:
@azazeal not green, but I blame wrong calculation of the actual diff 😛

azazeal (changes requested):
I don't see how two sentinel errors, both named `ErrNoStorageConfigured`, with the same error message, are helpful.

There's a few places where the following pattern takes place:

```go
something, err := doSomething()
if errors.Is(err, storage.ErrNotFound):
	return nil, fmt.Errorf("failed doing something: %w", ErrNotFound)
}

return nil, fmt.Errorf("failed doing something: %w", err)
```

I understand the intention is to return the `tpm` error here, instead of the `storage` one, I just don't see why this is necessary.

A sentinel error, `ErrNotConfigured` in the `storage` package, should be just enough, I think. After all, both the `storage` and `tpm` packages are public.

hslatman:
The reason I'd like the sentinel to be exposed from both packages, is so that a user of the `tpm` package doesn't need to import `storage` if they wanted to check for this specific case. 

I'll make the one in `tpm` an alias for the one from `storage`. That should result in the same behavior without having to put the switch/if in most places. 

Same approach can probably work for the existing sentinel errors; will check those too, and follow up in a new PR for those.

tpm/tss2/signer_test.go
```diff
@@ -55,7 +55,7 @@ var defaultKeyParamsRSAPSS = tpm2.Public{
 	},
 }
 
-func assertMaybeError(t assert.TestingT, err error, msgAndArgs ...interface{}) bool {
+func assertMaybeError(t assert.TestingT, err error, msgAndArgs ...any) bool {
 	return true
```
azazeal, line 59: Just caught this. Is this supposed to always return `true`?

tpm/tpm_test.go
```diff
@@ -74,3 +76,9 @@ func Test_close(t *testing.T) {
 	closeTPM(context.Background(), newCloseErrorTPM(t), &closeErr)
 	require.EqualError(t, closeErr, "failed closing attest.TPM: closeErr") // attest.TPM is backed by the closeSimulator
 }
+
+func TestTPMNoStorageConfiguredError(t *testing.T) {
+	err := ErrNoStorageConfigured
+	require.ErrorIs(t, err, ErrNoStorageConfigured)
```
azazeal, line 82: This first check is probably not required.

tpm/tss2/signer_test.go
azazeal: Will investigate further. Not your code.

tpm/tpm_test.go
hslatman: eb308a453240d2ec0ea3665991cd5935f5113ab0

azazeal (approved):

PR #580, opened 2024-08-29
Add method to obtain TPM capabilities

joshdrake:
This PR adds support for interrogating the TPM capabilities. The capabilities currently reported are the supported algorithms of the TPM.

As well, the TPMKMS will now:
- ensure that the TPM supports the algorithms for the requested signature algorithm
- default to the most appropriate supported signature algorithm based on supported algorithms

Future improvements may also ensure that the TPM supports the relevant hash/curve functions during sign operations.

💔Thank you!


tpm/info.go
```diff
@@ -14,13 +15,16 @@ import (
...
-	Interface       Interface       `json:"interface"`
-	Manufacturer    Manufacturer    `json:"manufacturer"`
-	VendorInfo      string          `json:"vendorInfo,omitempty"`
-	FirmwareVersion FirmwareVersion `json:"firmwareVersion,omitempty"`
+	Version             Version         `json:"version"`
+	Interface           Interface       `json:"interface"`
+	Manufacturer        Manufacturer    `json:"manufacturer"`
+	VendorInfo          string          `json:"vendorInfo,omitempty"`
+	FirmwareVersion     FirmwareVersion `json:"firmwareVersion,omitempty"`
+	SupportedAlgorithms []Algorithm     `json:"supportedAlgorithms"`
```
hslatman, line 23: Not an illogical place to put it, but I think exporting a new method on `TPM` will be better in the long term. It can be marked experimental, so that we can change the name, remove it later, etc. if we decide that's the best way forward.

joshdrake:
@hslatman Any examples on how I would go about testing the `SupportedAlgorithms` implementation? Can I use the simulator testing bits in conjunction with legacy `tpm2` package?

kms/tpmkms/tpmkms.go
```diff
@@ -73,21 +74,34 @@ type TPMKMS struct {
...
+var preferredAlgorithm = []apiv1.SignatureAlgorithm{
+	apiv1.ECDSAWithSHA384,
+	apiv1.ECDSAWithSHA512,
+	apiv1.ECDSAWithSHA256,
+	apiv1.SHA512WithRSA,
+	apiv1.SHA384WithRSA,
+	apiv1.SHA256WithRSA,
+	apiv1.SHA512WithRSAPSS,
+	apiv1.SHA384WithRSAPSS,
+	apiv1.SHA256WithRSAPSS,
```
joshdrake, line 104: I considered just having the keys of `signatureAlgorithmMapping` be the preferred order, but that seemed a little too terse. Happy to take another approach here.

kms/tpmkms/tpmkms.go
```diff
@@ -326,9 +340,30 @@ func (k *TPMKMS) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyRespons
...
+		}
+	} else {
+		v, ok = signatureAlgorithmMapping[req.SignatureAlgorithm]
+		if !ok {
+			return nil, fmt.Errorf("TPMKMS does not support signature algorithm %q", req.SignatureAlgorithm)
+		}
+
+		if !caps.SupportsAlgorithms(v.Requires...) {
+			return nil, fmt.Errorf("signature algorithm is %q not supported by TPM", req.SignatureAlgorithm)
+		}
```
joshdrake, line 366: This is additional behavior on top of the new "preferred" algorithm approach; it will verify whether the requested signature algorithm is supported by the TPM.

joshdrake: I thought this made sense, but if it's too aggressive of a change for right now, I'll remove.

tpm/caps.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Capabilities returns the capabilities of the TPM
+//
+// # Current supports enumerating the supported TPM algorithms
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
+// release.
+func (t *TPM) GetCapabilities(ctx context.Context) (caps *Capabilities, err error) {
```
hslatman, line 38:
```suggestion
// Capabilities returns the capabilities of the TPM
//
// Currently supports enumerating the supported TPM algorithms
//
// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
// release.
func (t *TPM) Capabilities(ctx context.Context) (caps *Capabilities, err error) {
```

tpm/caps.go
```diff
@@ -0,0 +1,73 @@
...
+	caps = &Capabilities{}
+
```
hslatman, line 40: You can make it store the capabilities for a future call, and return early if already obtained from the TPM. For a given TPM it doesn't change over time. It'll save a few tx/rx operations, especially if it sits in the path of creating keys. `tpm.Info` does something similar.

tpm/caps.go
```diff
@@ -0,0 +1,73 @@
...
+	if err = t.open(goTPMCall(ctx)); err != nil {
+		return nil, fmt.Errorf("failed opening TPM: %w", err)
+	}
+	defer closeTPM(ctx, t, &err)
+
+	current := tpm2.AlgUnknown // 0x0000, first property
+
+	for {
+		var (
+			data []interface{}
```
hslatman, line 50:
```suggestion
			data []any
```

tpm/caps.go
```diff
@@ -0,0 +1,73 @@
...
+			more bool
+		)
+
+		if data, more, err = tpm2.GetCapability(t.rwc, tpm2.CapabilityAlgs, 1, uint32(current)); err != nil {
+			return nil, fmt.Errorf("error getting algorithms capability: %w", err)
+		}
```
hslatman, line 56: This allows removing the var declaration before. Declaring the `more` within the `for` looks surprising.
```suggestion
			more bool
		)

		data, more, err := tpm2.GetCapability(t.rwc, tpm2.CapabilityAlgs, 1, uint32(current)); 
		if err != nil {
			return nil, fmt.Errorf("error getting algorithms capability: %w", err)
		}
```

tpm/caps.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/google/go-tpm/legacy/tpm2"
+	"go.step.sm/crypto/tpm/algorithm"
+)
+
+type Capabilities struct {
+	Algorithms []algorithm.Algorithm
+}
+
+func (c *Capabilities) SupportsAlgorithms(algs ...tpm2.Algorithm) bool {
```
hslatman, line 16: This will require a user of the package to also import `github.com/google/go-tpm/legacy/tpm2` explicitly. Better to prevent that by using our own type or alias, so that it's easier to guarantee the public API. I believe `algorithm.Algorithm` would already be suitable for it.

tpm/algorithm/algorithm.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return true
+}
+
+func init() {
+	// https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part2_Structures_pub.pdf
+	algs = map[tpm2.Algorithm]string{
+		// object types
+		tpm2.AlgRSA: "RSA",
+		tpm2.AlgECC: "ECC",
```
hslatman, line 47: This doesn't need to go into an `init`, but can be  put where the `algs` is initalized.

tpm/algorithm/algorithm.go
```diff
@@ -0,0 +1,98 @@
...
+	return algs[tpm2.Algorithm(int(a))]
+}
+
+func (a Algorithm) MarshalJSON() ([]byte, error) {
+	return json.Marshal(a.String())
+}
+
+type AlgorithmSlice struct {
+	algs []Algorithm
+}
```
hslatman, line 26: This possibly can be made a type alias. Depending on usage, could be named just `Algorithms`.

tpm/algorithm/algorithm.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+var (
+	algs map[tpm2.Algorithm]string
+)
+
+type Algorithm tpm2.Algorithm
+
+func (a Algorithm) String() string {
+	return algs[tpm2.Algorithm(int(a))]
```
hslatman, line 17: Add a guard for the case it doesn't exist in `algs`.

tpm/algorithm/algorithm.go
```diff
@@ -0,0 +1,98 @@
...
+func (a Algorithm) MarshalJSON() ([]byte, error) {
+	return json.Marshal(a.String())
+}
+
+type AlgorithmSlice struct {
+	algs []Algorithm
+}
+
+func (s *AlgorithmSlice) Supports(algs ...Algorithm) bool {
+	if len(algs) == 0 {
```
hslatman, line 29:
```suggestion
func (s *AlgorithmSlice) Contains(algs ...Algorithm) bool {
	if len(algs) == 0 {
```

tpm/algorithm/algorithm.go
```diff
@@ -0,0 +1,98 @@
...
+	algs = map[tpm2.Algorithm]string{
+		// object types
+		tpm2.AlgRSA: "RSA",
+		tpm2.AlgECC: "ECC",
+
+		// encryption algs
+		tpm2.AlgRSAES: "RSAES",
+
+		// block ciphers
+		0x0003:            "3DES",
```
hslatman, line 53: All that are not defined in `tpm2` can be made into a `tpm2.Algorithm` like this:
```suggestion
		tpm2.Algorithm(0x0003):            "3DES",
```

kms/tpmkms/tpmkms.go
```diff
@@ -73,21 +74,34 @@ type TPMKMS struct {
 }
 
 type algorithmAttributes struct {
-	Type  string
-	Curve int
+	Type     string
+	Curve    int
+	Requires []tpm2.Algorithm
```
hslatman, line 79: See remark about having to import `tpm2`

kms/tpmkms/tpmkms.go
hslatman: I think the functionality is nice, but the order defined here might be too opinionated and possibly artificial to impose here, especially when it comes to the order of hashing algorithms. Given the hash calculations are performed by the TPM, higher numbers are likely to have an impact on performance that is more noticeable than when performed on a CPU.

kms/tpmkms/tpmkms.go
```diff
@@ -326,9 +340,32 @@ func (k *TPMKMS) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyRespons
 		return nil, fmt.Errorf("failed parsing %q: %w", req.Name, err)
 	}
 
-	v, ok := signatureAlgorithmMapping[req.SignatureAlgorithm]
-	if !ok {
-		return nil, fmt.Errorf("TPMKMS does not support signature algorithm %q", req.SignatureAlgorithm)
+	ctx := context.Background()
+	caps, err := k.tpm.GetCapabilities(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("could not get TPM capabilities %w", err)
```
hslatman, line 346:
```suggestion
		return nil, fmt.Errorf("could not get TPM capabilities: %w", err)
```

kms/tpmkms/tpmkms.go
```diff
@@ -326,9 +340,32 @@ func (k *TPMKMS) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyRespons
...
+			}
+		}
+	} else {
+		v, ok = signatureAlgorithmMapping[req.SignatureAlgorithm]
+		if !ok {
+			return nil, fmt.Errorf("TPMKMS does not support signature algorithm %q", req.SignatureAlgorithm)
+		}
+
+		if !caps.SupportsAlgorithms(v.Requires...) {
+			return nil, fmt.Errorf("signature algorithm is %q not supported by the TPM device", req.SignatureAlgorithm)
```
hslatman, line 367:
```suggestion
			return nil, fmt.Errorf("signature algorithm %q is not supported by the TPM device", req.SignatureAlgorithm)
```

kms/tpmkms/tpmkms.go
hslatman: The part checking algorithm support is great, as that'll allow us to return a more descriptive error message, instead of relying on a possibly more cryptic TPM error code. That's been on my wishlist for a while, in fact 😅 We must ensure that it's working as expected with actual TPMs, as it could mean we'd be blocking more operations than before. But I think it's fairly low risk at this time.

If possible, I would make selection of the preferred algorithm something that becomes configurable, as it's fairly opinionated at this time, and thus becomes opt-in. It could be on the `TPMKMS` level when instantiating the KMS, when doing key creation operations, or both.

kms/tpmkms/tpmkms.go
```diff
@@ -326,9 +340,32 @@ func (k *TPMKMS) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyRespons
...
+	var (
+		v  algorithmAttributes
+		ok bool
+	)
+	if !properties.ak && req.SignatureAlgorithm == apiv1.UnspecifiedSignAlgorithm {
+		for _, alg := range preferredAlgorithm {
+			v, ok = signatureAlgorithmMapping[alg]
+			if caps.SupportsAlgorithms(v.Requires...) {
+				break
+			}
```
hslatman, line 358: This doesn't seem to be checking `ok`. Maybe `v` is zero value in this case?

kms/tpmkms/tpmkms.go
```diff
@@ -326,9 +340,32 @@ func (k *TPMKMS) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyRespons
...
+	ctx := context.Background()
+	caps, err := k.tpm.GetCapabilities(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("could not get TPM capabilities %w", err)
+	}
+
+	var (
+		v  algorithmAttributes
+		ok bool
+	)
```
hslatman, line 352: Defining these here seems unnecessary if you make them assignments in both parts below.

hslatman:
> @hslatman Any examples on how I would go about testing the `SupportedAlgorithms` implementation? Can I use the simulator testing bits in conjunction with legacy `tpm2` package?

I take it you mean the `tpm.Capabilities` method? It should be possible to use the simulator for a basic test. That test would go into `tpm_simulator_test.go`. Setup would look similar to the one for getting the TPM info: https://github.com/smallstep/crypto/blob/master/tpm/tpm_simulator_test.go#L60-L78.

The test would be limited to what the TPM simulator returns in terms of supported algorithms, which would always be the same, unless recompiled per test case, which doesn't sound enticing to me 😅 AFAIK there's no way to reconfigure the TPM simulator at runtime. To have more test variability we could look into mocking TPM responses on the `io.ReadWriterCloser` level, possibly fed by responses from an actual TPM. For relatively static things like the capabilities, it'll probably be OK.

tpm/caps.go
joshdrake: 665a65a

kms/tpmkms/tpmkms.go
joshdrake: d883afe

kms/tpmkms/tpmkms.go
joshdrake: d883afe

kms/tpmkms/tpmkms.go
joshdrake: 83d7282

kms/tpmkms/tpmkms.go
joshdrake: 83d7282

tpm/algorithm/algorithm.go
joshdrake: I replaced this type and `Supports`/`Contains` with `SupportsAlgorithms` in the `tpm` package.

tpm/algorithm/algorithm.go
joshdrake: d7e1326

tpm/caps.go
joshdrake: 657022a

tpm/algorithm/algorithm.go
joshdrake: 657022a

kms/tpmkms/tpmkms.go
joshdrake: 657022a

kms/tpmkms/tpmkms.go
joshdrake: `v` is used further down, I could refactor more of this code if you feel strongly.

hslatman (dismissed):
Nice 👍 

Just a linting issue left

tpm/tpm.go
```diff
@@ -74,6 +75,15 @@ func WithDisableDownload() NewTPMOption {
 	}
 }
 
+// WithCapabilities explicits sets the capabilities rather
+// than acquiring them from the TPM directly.
+func WithCapabilities(caps *Capabilities) NewTPMOption {
```
hslatman, line 80: Can it be made available in tests only by adding it to `tpm_test.go`? I'm not sure if we want to support it as an actual option. And maybe the property in the options can be named `testCapabilities`, or something like that, to make it explicit.

kms/tpmkms/tpmkms_simulator_test.go
```diff
@@ -59,10 +59,19 @@ func withKey(name string) newSimulatedTPMOption {
 	}
 }
 
-func newSimulatedTPM(t *testing.T, opts ...newSimulatedTPMOption) *tpmp.TPM {
+func newSimulatedTPM(t *testing.T, caps *tpmp.Capabilities, opts ...newSimulatedTPMOption) *tpmp.TPM {
```
hslatman, line 62: Can you turn the capabilities into a `newSimulatedTPMOption`, like `withAK` and `withKey`, only not operating on the `tpm.TPM` instance? Then it can remain variadic, and won't require changing it in the tests where you're now passing `nil`. It'll still need the special case to properly create the TPM instance, so there's still some logic needed to switch between "initialization" and "preparation" options.

kms/tpmkms/tpmkms_simulator_test.go
```diff
@@ -433,8 +442,62 @@ func TestTPMKMS_CreateKey(t *testing.T) {
 	}
 }
 
+func TestTPMKMS_CreateKey_Capabilities(t *testing.T) {
+	tpmWithNoCaps := newSimulatedTPM(t, &tpmp.Capabilities{}, withAK("ak1"))
```
hslatman, line 446: The `withAK` option doesn't seem to be used in the test, as it's creating keys; not attesting their creation.

kms/tpmkms/tpmkms.go
```diff
@@ -326,9 +354,36 @@ func (k *TPMKMS) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyRespons
...
+	caps, err := k.tpm.GetCapabilities(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("could not get TPM capabilities: %w", err)
+	}
+
+	var (
+		v  algorithmAttributes
+		ok bool
+	)
+	if !properties.ak && req.SignatureAlgorithm == apiv1.UnspecifiedSignAlgorithm && len(preferredSignatureAlgorithms) > 0 {
```
hslatman, line 367: Just saw the usage of `withAK` and was thinking how that could succeed, and then found that the algorithm is ignored here. I think that's fine for now, but we may want to add the capability check for AK creation too at some point.

tpm/caps.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/google/go-tpm/legacy/tpm2"
+	"go.step.sm/crypto/tpm/algorithm"
+)
+
+type Capabilities struct {
+	Algorithms []algorithm.Algorithm
+}
+
+func (c *Capabilities) SupportsAlgorithms(algs ...algorithm.Algorithm) bool {
```
hslatman, line 16: Did you want to change it to singular, or did you change your mind? Don't forget the doc on the struct and the method.

kms/tpmkms/tpmkms_simulator_test.go
joshdrake: Yea, I'll work on this more - I wanted to do that originally, but I think I'll have to change how options work between this package and the `tpm` package to be able to actually modify the capabilities.

hslatman: I think it can be done by changing the `newSimulatedTPMOption` to also return a type or bool (or to rework that into a struct), and to use that to differentiate between initialization and preparation options.

tpm/caps.go
joshdrake: [d378c7d](https://github.com/smallstep/crypto/pull/580/commits/d378c7d363e5a5c9691dcae6e3184863058fc287)

kms/tpmkms/tpmkms_simulator_test.go
joshdrake: @hslatman  Unless I'm not following you, I can't change this behavior without either exporting _something_ from the `tpm` package to control the capabilities while under test (eg: either a helper `Option` setter, the `options` struct and field, etc.). I've removed that test code for now: what do you think about either mocking the `tpm` package itself or adding support to the `Simulator` to override the capabilities, eg (in the `simulator` package):

```
// the packged command buffer for CmdGetCapability (0x17A)
var getCapabilityAlgs = []byte{128, 1, 0, 0, 0, 22, 0, 0, 1, 122, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1}

func (s *WrappingSimulator) Write(p []byte) (int, error) {
	if bytes.Equal(p, getCapabilityAlgs) {
		// ...
	}
	return s.wrapped.Write(p)
}
```

joshdrake: Also, if you don't mind, I'll keep working on better tests for this after merging, so we can unblock work depending on these changes.

hslatman: I've added the test you had back in https://github.com/smallstep/crypto/pull/590 and merged that into this branch. You're right we'll need that `WithCapabilities` method exported; didn't think about the package boundary. I've marked it experimental, and noted that it's primarily used for testing. 

Some form of emulating responses from a TPM might be useful for future tests, and it would be great to have that available in plain Go, but I think it would be best in a new package, and not (partially) reliant on the C simulator. I know Panos has looked a bit into generating Go code from the spec, so that might be a way to do it.

hslatman (approved):

PR #600, opened 2024-09-26
Add support for using mackms keys with no tags

### LINKED ISSUE #595: Using the mackms package to load keys that do not have a tag attribute set
### What would you like to be added?

Hey there!

I really like the `mackms` package- it makes working with the keychain from Go really pleasant. 🎉 

One issue I'm running into is that the current URI structure [sets a default `tag`](https://github.com/smallstep/crypto/blob/master/kms/mackms/mackms.go#L1015-L1018) attribute if one is not provided. This makes it not possible to work with existing keys that do not have a tag on them today.

It doesn't appear possible to explicitly tell `mackms` to _not_ add a tag attribute to the `SecItemCopyMatching` queries.

Is this something that you'd be open to considering? I know that changing the existing default behavior wouldn't be backwards compatible. Just tossing out an option, maybe an approach for supporting `null*` style parameters in the URI, e.g. `label=my-name;hash=ccb792f9d9a1262bfb814a339876f825bdba1261;nulltag=true` might work for explicitly unsetting attributes.

Note that this might require adding more conditionals like [this one](https://github.com/smallstep/crypto/blob/master/kms/mackms/mackms.go#L726-L733) so that tags are truly excluded from the search dictionary.

### Why this is needed

This will make it easier to work with existing keychain items that may not have particular attributes like tags.

I'm happy to help with this if there's an approach you like, but either way I appreciate y'all considering this! Keep up the great work.

maraino:
### Description

This commit allows to create, get, sign, and other operations using keys without a tag.

By default, the default tag used by the mackms package is `com.smallstep.crypto`. If we want to create or get a key using the URI `mackms:label=test`, the `mackms` package will assume that the default tag is being used.

The default tag can be changed using the tag parameter in the URI, e.g., `mackms:label=test;tag=my-tag`. But if we want not to use a tag, we need to provide the tag parameter empty, e,g., `mackms:label=test;tag=`. 

Programs using the `mackms` package can also set the `mackms.DefaultTag` to any other value, if it is set to empty, the created keys will not have a tag.

Fixes #595

### DIFF
--- a/kms/mackms/mackms.go
+++ b/kms/mackms/mackms.go
@@ -95,6 +95,7 @@ var signatureAlgorithmMapping = map[apiv1.SignatureAlgorithm]algorithmAttributes
 // CreateKey methods can create keys with the following URIs:
 //   - mackms:label=my-name
 //   - mackms:label=my-name;tag=com.smallstep.crypto
+//   - mackms;label=my-name;tag=
 //   - mackms;label=my-name;se=true;bio=true
 //
 // GetPublicKey and CreateSigner accepts the above URIs as well as the following
@@ -107,7 +108,8 @@ var signatureAlgorithmMapping = map[apiv1.SignatureAlgorithm]algorithmAttributes
 //     represents the key name. You will be able to see the keys in the Keychain,
 //     looking for the value.
 //   - "tag" corresponds with kSecAttrApplicationTag. It defaults to
-//     com.smallstep.crypto.
+//     com.smallstep.crypto. If tag is an empty string ("tag="), the attribute
+//     will not be set.
 //   - "se" is a boolean. If set to true, it will store the key in the
 //     Secure Enclave. This option requires the application to be code-signed
 //     with the appropriate entitlements.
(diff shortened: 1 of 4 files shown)


dopey (approved):

maraino:
@hslatman, Can you look at it too?

hslatman (commented):
@maraino logic looks OK.

It would be nice to update (some of) the doc blocks describing the empty tag usage. Maybe not a full example URI, but at least a short blurb saying an empty tag is allowed, and results in creating / searching / deleting keys without a tag being set.

I think it's a small enough of a use case that just the empty tag is OK, but we could opt to (also) support a specific sentinel value for the new behavior, i.e. `tag=<null>`. Don't have a strong preference, though.

maraino:
@hslatman I've added an example in the docs (6844893)

Yesterday, I tested setting an empty string tag, but macOS treats it like not setting any. In any case, I decided to add the if conditions to make it more clear.

hslatman (approved):

PR #631, opened 2024-11-14
Allow to retry with no tags when loading privatekeys

maraino:
### Description

This function allows you to retry the load of a key without tags if the original URL didn't have one.

This is used only in GetPublicKey and CreateSigner. It is not used in SearchKeys as duplicate keys must be removed.

cc: @joshdrake

### DIFF
--- a/kms/mackms/mackms.go
+++ b/kms/mackms/mackms.go
@@ -65,10 +65,8 @@ type keyAttributes struct {
 // would return the tag empty if it was set using the default value.
 func (k *keyAttributes) retryAttributes() *keyAttributes {
 	if !k.retry {
-		fmt.Println("no retry")
 		return nil
 	}
-	fmt.Println("retry")
 	return &keyAttributes{
 		label: k.label,
 		hash:  k.hash,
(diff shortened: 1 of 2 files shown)


kms/mackms/mackms.go
```diff
@@ -52,12 +52,30 @@ type keyAttributes struct {
...
+// retryAttributes returns the original URI attributes used to get a private
+// key, but only if they are different that the ones set. It will return nil, if
+// they are the same. The only attribute that can change is the tag. This method
+// would return the tag empty if it was set using the default value.
+func (k *keyAttributes) retryAttributes() *keyAttributes {
+	if !k.retry {
+		fmt.Println("no retry")
+		return nil
+	}
+	fmt.Println("retry")
```
joshdrake, line 71:
```suggestion
```

joshdrake (changes requested):
One errant debugging statement, otherwise lgtm.

kms/mackms/mackms.go
maraino: Fixed with d01bccd

joshdrake (approved):

PR #656, opened 2024-12-18
Add support for imported keys

### LINKED ISSUE #655: Add support for imported keys on yubikey kms
### Description

`step-ca` or `step-kms-plugin` fails to sign with a key imported into the yubikey. The cause of this is because we attest the slot to get the public key. We also support the workaround of getting the certificate in the same slot. But we can easily support imported keys without the workaround if we use KeyInfo to get the public key, it would be something something like this:
```go
func (k *YubiKey) getPublicKey(slot piv.Slot) (crypto.PublicKey, error) {
	ki, err := k.yk.KeyInfo(slot)
	if err != nil {
		return nil, err
	}
	// not use if this can be nil
	if ki.PublicKey == nil {
		 return nil, errors.New("error retrieving public key")
	}
	return ki.PublicKey, nil
}
```

KeyInfo is only supported on YubiKeys with a version >= 5.3.0, so we might need to fallback to Attest and Certificate.

maraino:
### Description

This commit adds support for imported keys in YubiKey KMS. Now, we attempt to retrieve a key using go-piv's KeyInfo that supports both imported and generated keys. This functionality is only available from YubiKey firmware  5.3.0. We will fallback to Attest and Certificate methods that are available in older versions.

Fixes #655

### DIFF
--- a/kms/yubikey/yubikey.go
+++ b/kms/yubikey/yubikey.go
@@ -42,6 +42,7 @@ type pivKey interface {
 	Certificate(slot piv.Slot) (*x509.Certificate, error)
 	SetCertificate(key []byte, slot piv.Slot, cert *x509.Certificate) error
 	GenerateKey(key []byte, slot piv.Slot, opts piv.Key) (crypto.PublicKey, error)
+	KeyInfo(slot piv.Slot) (piv.KeyInfo, error)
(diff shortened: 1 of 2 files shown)


hslatman (approved):
Looks OK!

PR #689, opened 2025-01-28
Extend jose.NewSigner support for other types

maraino:
This commit adds support for custom `crypto.Signer` types for signing tokens. With this, one can use the signer implemented in a KMS or even `step` + `step-kms-plugin` to sign JWTs.

### DIFF
--- a/jose/parse.go
+++ b/jose/parse.go
@@ -358,6 +358,19 @@ func guessJWKAlgorithm(ctx *context, jwk *JSONWebKey) {
 	}
 }
 
+func guessOpaqueSigner(key crypto.PrivateKey) crypto.PrivateKey {
+	switch k := key.(type) {
+	case []byte, *ecdsa.PrivateKey, *rsa.PrivateKey, ed25519.PrivateKey:
+		return key
+	case x25519.PrivateKey:
+		return X25519Signer(k)
+	case crypto.Signer:
+		return NewOpaqueSigner(k)
+	default:
+		return key
+	}
+}
+
 // guessSignatureAlgorithm returns the signature algorithm for a given private key.
 func guessSignatureAlgorithm(key crypto.PrivateKey) SignatureAlgorithm {
 	switch k := key.(type) {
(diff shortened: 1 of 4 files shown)


jose/parse.go
```diff
@@ -358,6 +358,19 @@ func guessJWKAlgorithm(ctx *context, jwk *JSONWebKey) {
 	}
 }
 
+func guessOpaqueSigner(key crypto.PrivateKey) crypto.PrivateKey {
```
hslatman, line 361: Should this be called `guessSigner`, since it can also return other types of signers, specifically the `X25519Signer` (and will leave others untouched)?

jose/parse_test.go
```diff
@@ -33,6 +36,24 @@ const (
...
+	if w.Signer == nil {
+		return nil
+	}
+	return w.Signer.Public()
+}
+
+func (w wrapSigner) Sign(r io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error) {
+	if w.Signer == nil {
+		return nil, errors.New("not implemented")
+	}
```
hslatman, line 53: Since `wrapSigner` is only used in tests, you could make it take `t *testing.T`, and fail immediately when `Signer` is nil. Then you can remove some test cases that are only testing this test struct's functioning.

jose/parse.go
maraino: `X25519Signer` is an implementation of the `OpaqueSigner` interface.

maraino: The method intends to return an OpaqueSigner only if necessary.

jose/parse_test.go
maraino: There are specific cases where I don't want the test to fail. For example, when I create an opaque signer with `NewOpaqueSigner(wrapSigner{})`, this method will end up calling `signer.Public()`, I don't care about the output, but I don't want the method to fail, I want only that the opaqueSigner is not able to recognize the type of the key.

hslatman (approved):

PR #695, opened 2025-02-07
Replace conflicting nssdb objects

areed:
When importing certificates and keys into an NSS db replace any with a conflicting CKA_ID, which holds the certs subject key id.

Add new DeleteCertificatesByName command to clean up certificates and their keys by nickname.

Also fix bug with generating initialization vector for aes256-cbc. Also fix pk1sign test that allowed bug to go undetected.

💔Thank you!


nssdb/certificate_test.go
```diff
@@ -167,12 +181,16 @@ func TestNSSDB_Import(t *testing.T) {
 			if err != nil {
 				t.Skip("pk1sign not installed")
 			}
+
 			db, dir := v.connectDir(t)
+			require.NoError(t, db.Reset(ctx))
```
areed, line 186: This is why the `pk1sign` test was passing for the IV without the 040e prefix - because it was testing with a key imported with `pk12util` not the one imported by this package.

nssdb/certificate.go
```diff
@@ -87,6 +87,16 @@ func (db *NSSDB) AddCertificate(ctx context.Context, cert *x509.Certificate, nam
...
 
+	matches, err := db.findByAttr(ctx, CKO_CERTIFICATE, "CKA_ID", cert.SubjectKeyId)
+	if err != nil {
+		return 0, 0, fmt.Errorf("find cka id conflicts: %w", err)
+	}
+	for _, id := range matches {
+		if err := db.DeleteCertificate(ctx, id); err != nil {
+			return 0, 0, fmt.Errorf("delete conflicting certificate %d: %w", id, err)
+		}
+	}
```
maraino, line 98: It might make sense for our use case, but is this really necessary in NSS? The subjectKeyId only indicates that the key used is the same. I think we should be able to store a certificate with the same key. 

To me, it might make sense to replace them if the certificates have the same label, although, not allowing that, as we do in the KMSs is a valid option too.

Replacing a different certificate with the same subjectKeyId should be a decision of the application using this package, not by the package.

nssdb/keys.go
```diff
@@ -182,14 +195,27 @@ func (db *NSSDB) AddPrivateKey(ctx context.Context, privKey *ecdsa.PrivateKey, n
...
-	pubKeyObj, err := ecPubKeyToObject(pubKey, id)
+	if len(ckaID) > 0 {
+		matches, err := db.findByAttr(ctx, CKO_PUBLIC_KEY, "CKA_ID", ckaID)
+		if err != nil {
+			return 0, fmt.Errorf("find cka id conflicts: %w", err)
+		}
+		for _, id := range matches {
+			if err := db.DeleteObjectPublic(ctx, id); err != nil {
+				return 0, fmt.Errorf("delete conflicting public key %d: %w", id, err)
+			}
```
maraino, line 214: Per before, replacing objects with the same id or label might make sense, but it's also an option to fail, and let the user delete it if it already exists. We do that in the KMS, even in `step`, where the decision is taken by the user, either typing yes or passing the `--force` flag.

I think we should consider what the cli utils that are used to manage NSS databases do by default.

maraino (changes requested):
Deleting objects is always dangerous, I think we should consider what the NSS tools do in those cases. 

To me, it should be an option of the user, in this case, by the application using this package. But it would be ok, if that is the behavior used by the default tools used to manage those databases.

nssdb/nssdb.go
```diff
@@ -492,3 +492,48 @@ func (db *NSSDB) getObjectID(ctx context.Context) (uint32, error) {
...
+		q := fmt.Sprintf("SELECT id FROM nssPrivate WHERE %s = ? AND a0 = X'00000003' AND id IS NOT NULL", col)
+		rows, err = db.Key.QueryContext(ctx, q, val)
+	case CKO_PUBLIC_KEY:
+		//nolint:gosec // trusted column name
+		q := fmt.Sprintf("SELECT id FROM nssPublic WHERE %s = ? AND a0 = X'00000002' AND id IS NOT NULL", col)
+		rows, err = db.Cert.QueryContext(ctx, q, val)
+	case CKO_CERTIFICATE:
+		//nolint:gosec // trusted column name
+		q := fmt.Sprintf("SELECT id FROM nssPublic WHERE %s = ? AND a0 = X'00000001' AND a80 = X'00000000' AND id IS NOT NULL", col)
+		rows, err = db.Cert.QueryContext(ctx, q, val)
```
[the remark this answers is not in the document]
areed: I've updated to use constants instead of literals in b66b3f3. The `a0` and `a80` columns are used in the db.

nssdb/certificate.go
```diff
@@ -87,6 +87,16 @@ func (db *NSSDB) AddCertificate(ctx context.Context, cert *x509.Certificate, nam
 		return 0, 0, err
 	}
 
+	matches, err := db.findByAttr(ctx, CKO_CERTIFICATE, "CKA_ID", cert.SubjectKeyId)
+	if err != nil {
+		return 0, 0, fmt.Errorf("find cka id conflicts: %w", err)
```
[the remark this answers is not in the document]
areed: b66b3f3

nssdb/certificate.go
areed: I copied `pk12util` where importing with the same subject key id deletes the existing certificate and importing with the same label or nickname leaves the original.

nssdb/keys.go
areed: Keys are a bit different than certificates. `pk12util` replaces the certificate with the same CKA_ID but for keys it leaves the old ones in place. This package deletes and re-creates so the keys will have updated ids but the attributes should be the same.

nssdb/nssdb.go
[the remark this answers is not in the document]
maraino: I suppose those column names won't change.

maraino (approved):
Although the behavior of deleting objects is not ideal, I think it should be an option in the app using this, but as it's the default behavior in pk12util, let's go with this for now.

Issue #241, opened 2023-05-19
Provide a means for KMSs to not fail hard when their initialization fails


hslatman:
### What would you like to be added

An option to initialize a KMS ignoring initialization errors related to side effects.

### Why this is needed

When a KMS is initialized, it can happen that a system the KMS depends on is not available at the time of initialization, but could be at time of usage. Currently the behavior results in returning an error immediately on failure, which usually results in a failure to continue operating. For some use cases that's fine, such as when a CLI is used to interact with a KMS. It's different when a server process is starting up, and the KMS is not available. The server won't start in that case, unless logic is implemented for handling the specific error.

A solution could be to return a sentinel error that the caller can check for and then decide what to do with it. Another option is to provide options to initialization, affecting the initialization logic. Being able to somehow check the "health" of a KMS might also be useful, similar to how one can ping a DB from inside a process to see if it's still available.

PR #702, opened 2025-02-13
Add `CreateDecrypter` support to AWS KMS

hslatman:


kms/awskms/decrypter.go
```diff
@@ -0,0 +1,172 @@
...
+		rsaOpts = o
+	case *rsa.PKCS1v15DecryptOptions:
+		return "", errors.New("awskms does not support PKCS #1 v1.5 decryption")
+	default:
+		return "", fmt.Errorf("invalid decrypter options type %T", opts)
+	}
+
+	switch pub.Size() {
+	default:
+		return "", fmt.Errorf("awskms does not support RSA public key size %d", pub.Size())
```
sduc, line 136:
```suggestion

        bitSize := pub.Size() * 8
	switch bitSize {
	default:
		return "", fmt.Errorf("awskms does not support RSA public key size %d", bitSize)
```
`pub.Size()` is in bytes so the switch will always say we don't have the right key size.

hslatman: Thanks! 7c54eedf31a33db2627261138843e90df7fc1c82 🙂

hslatman:
I've just done a quick encrypt/decrypt test against AWS, and it looks OK 🙂

sduc:
> I've just done a quick encrypt/decrypt test against AWS with a patched `step-kms-plugin`, and it looks OK 🙂
> 
> ```
> $ step-kms-plugin encrypt --oaep --in message.txt --no-label --kms 'awskms:' 'awskms:key-id=<key-id>;region=<region>'
> // encrypted output in base64 format
> 
> $ step-kms-plugin decrypt --oaep --in encrypted.bin --no-label --kms 'awskms:' 'awskms:key-id=<key-id>;region=<region>'
> // decrypted output
> ```

Yup I confirm. The only requirement is to have an RSA key configured with `Key Usage:
Encrypt and decrypt` and not `Sign and Verify` :)

hslatman:
> Yup I confirm. The only requirement is to have an RSA key configured with `Key Usage: Encrypt and decrypt` and not `Sign and Verify` :)

Indeed. I've seen that before with GCP Cloud KMS, and is considered a good practice. In its current form we generally don't check the key capabilities in the KMS implementations before performing crypto operations. An invalid key will thus result in an error upon trying a decryption, whereas it may be an option to check the capabilities before trying to use the key for a purpose it's not configured for. Maybe we'll add something for that in the future 🙂 

For your use case you'll thus need the SCEP decrypter configured.

azazeal:
Is seems there's a lot of places where a new context pops out instead of being propagated. Is that intentional?

hslatman:
> Is seems there's a lot of places where a new context pops out instead of being propagated. Is that intentional?

These are the two I see:
- There's no `DecryptContext(ctx context.Context, ...)` in the stdlib, so it has to come from somewhere when calling a function that expects one when it's being called through the `crypto.Decrypter` interface.
- For `preloadKey` we maybe can provide an alternative, but I don't consider that part of the scope of this PR. There's an [open issue](https://github.com/smallstep/crypto/issues/241) to make initialization of a KMS lazy, and I suspect we may want to touch on `context.Context` usage there too.

maraino (approved):
lgtm

PR #703, opened 2025-02-14
Use  SHA-256 for subject key id on FIPS 140-3 mode

maraino:
### Description

This commit adds the fips util package, which reports whether the cryptography libraries are operating in FIPS 140-3 mode.

It also uses the 160 leftmost bits of SHA-256 to generate the Subject Key IDs.

Related:
- https://github.com/golang/go/issues/71746


hslatman:
Linter issues fixed in https://github.com/smallstep/crypto/pull/704.

x509util/utils_test.go
```diff
@@ -176,6 +182,38 @@ func Test_generateSubjectKeyID(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := generateSubjectKeyID(tt.args.pub)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("generateSubjectKeyID() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("generateSubjectKeyID() = %v, want %v", got, tt.want)
+			}
```
hslatman, line 212: You could use `require` here. It's used in some other places in `crypto` already.

x509util/utils_test.go
```diff
@@ -176,6 +182,38 @@ func Test_generateSubjectKeyID(t *testing.T) {
 	}
 }
 
+func Test_generateSubjectKeyID_fips(t *testing.T) {
+	if !fipsutil.Enabled() {
+		t.Skip("FIPS 140-3 mode is not enabled")
+	}
```
hslatman, line 188: Could `t.Setenv` help executing the tests at all times, but with the `GODEBUG` env var set for this test?

maraino: No, I've tried but GODEBUG cannot be changed after initializing, it will panic if you do it. See https://github.com/golang/go/blob/279da965329a74cd75320f15cb9672a282690ab7/src/crypto/fips140/fips140.go#L26-L28
```go
if currentlyEnabled != fips140.Enabled {
        panic("crypto/fips140: GODEBUG setting changed after program start")
}
```

jose/encrypt.go
```diff
@@ -104,7 +104,7 @@ func Decrypt(data []byte, opts ...Option) ([]byte, error) {
 		if data, err = enc.Decrypt(ctx.password); err == nil {
 			return data, nil
 		}
-		return nil, errors.New("failed to decrypt JWE: invalid password")
+		return nil, errors.New("failed to decrypt JWE: invalid password" + err.Error())
```
hslatman, line 107: Added a space, but other options might work too. No `fmt.Errorf`?
```suggestion
		return nil, errors.New("failed to decrypt JWE: invalid password " + err.Error())
```

x509util/utils_test.go
```diff
@@ -9,6 +9,9 @@ import (
 	"os"
 	"reflect"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.step.sm/crypto/fipsutil"
```
hslatman, line 14:
```suggestion
	"github.com/stretchr/testify/assert"
	
	"go.step.sm/crypto/fipsutil"
```

hslatman (approved):

PR #710, opened 2025-02-20
Use utf8 string for private key subject with non-printable characters

areed:
When importing a private key into an NSS database with a certificate, the certificate common name is used as the value of the CKA_SUBJECT attribute on the key object. This fixes the serialization to work when the CN contains characters that aren't valid for the asn1 printable string type.

💔Thank you!

### DIFF
--- a/nssdb/keys.go
+++ b/nssdb/keys.go
@@ -241,7 +241,7 @@ func (db *NSSDB) AddPublicKey(ctx context.Context, pubKey *ecdsa.PublicKey, ckaI
 	return pubKeyID, nil
 }
 
-var printableRx = regexp.MustCompile(`^[a-zA-Z0-9 '()+,-.\\:=?]*$`)
+var printableRx = regexp.MustCompile(`^[a-zA-Z0-9 '()+,\-.\\:=?]*$`)
 
 func isPrintable(s string) bool {
 	return printableRx.MatchString(s)
--- a/nssdb/keys.go
+++ b/nssdb/keys.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"regexp"
 
 	"golang.org/x/crypto/cryptobyte"
 )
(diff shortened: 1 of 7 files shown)


joshdrake (approved):

nssdb/keys.go
```diff
@@ -227,3 +240,9 @@ func (db *NSSDB) AddPublicKey(ctx context.Context, pubKey *ecdsa.PublicKey, ckaI
 
 	return pubKeyID, nil
 }
+
+var printableRx = regexp.MustCompile(`^[a-zA-Z0-9 '()+,\-.\\:=?]*$`)
+
+func isPrintable(s string) bool {
+	return printableRx.MatchString(s)
+}
```
maraino, line 248: Is this a limitation of nssdb? 

The [unicode](https://pkg.go.dev/unicode) package has a method for validating whether a run is printable [IsPrint(rune)](https://pkg.go.dev/unicode#IsPrint). Some more complicated limitations can be done using the [In(r rune, ranges ...*RangeTable)](https://pkg.go.dev/unicode#In) function, for example, to limit the runes to be in the `Latin` [RangeTable](https://pkg.go.dev/unicode#RangeTable), or you can also create your own.

areed: This refers to the asn.1 type of printable strings, which is more restrictive than the unicode IsPrint function. In a certificate where the common name contains only printable characters it's encoded with the asn1 tag for PrintableString. For anything else it gets the asn1 tag for UTF8String. (At least that's how it works for our certificates.)

maraino: Looping runes will perform better:
https://github.com/smallstep/crypto/blob/9fd47c3cda1afa9da212dd49dbe24622d1482b85/x509util/utils.go#L183-L213

Go ans1 loops around bytes using a similar method, this is for one byte https://github.com/golang/go/blob/f24b299df2896a4e8a80863dbb55a264f4b9bb68/src/encoding/asn1/asn1.go#L422-L444

areed: 9ce1394

maraino (approved):
lgtm

PR #730, opened 2025-03-20
Add support for tokens with no pins

maraino:
### Description

This commit sets the option LoginNotSupported if no pin is set. Servers using the p11-kit server protocol might not have a pin.

@hslatman should we do that? Or add an extra option? 
We don't have any other way to set a pin, so doing this I think is ok for now.

### DIFF
--- a/kms/pkcs11/pkcs11.go
+++ b/kms/pkcs11/pkcs11.go
@@ -121,6 +121,10 @@ func New(_ context.Context, opts apiv1.Options) (*PKCS11, error) {
 	if config.Pin == "" && opts.Pin != "" {
 		config.Pin = opts.Pin
 	}
+	// If no pin is set, assume that the token does not support login.
+	if config.Pin == "" {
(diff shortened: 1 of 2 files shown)


hslatman:
An additional option specifically for this behavior might be better, as the `LoginNotSupported` config controls the login flow in `crypto11`. I think performing a login with an empty PIN could still be useful, although I don't know if empty PINs are something that are supported generally.

maraino:
> An additional option specifically for this behavior might be better, as the `LoginNotSupported` config controls the login flow in `crypto11`. I think performing a login with an empty PIN could still be useful, although I don't know if empty PINs are something that are supported generally.

What do you think about ed0b190? It supports empty pins with `pin-value=`. 

Using no special arguments matches better with tooks like `pkcs11-tool` where it doesn't attempt a login if you don't add a pin.

hslatman (approved):

hslatman:
Yeah, this seems more explicit.

PR #744, opened 2025-04-01
Fix templates with policies with Go 1.24

### LINKED ISSUE #738: Update x509util.Certificate to manage policies
### Description

In Go 1.24, the field `PolicyIdentifiers` on `x509.Certificate` has changed, and only `Policies` is marshaled by default. See this in https://pkg.go.dev/crypto/x509#CreateCertificate:

The PolicyIdentifier and Policies fields can both be used to marshal certificate policy OIDs. By default, only the Policies is marshaled, but if the GODEBUG setting "x509usepolicies" has the value "0", the PolicyIdentifiers field will be marshaled instead of the Policies field. This changed in Go 1.24. The Policies field can be used to marshal policy OIDs which have components that are larger than 31 bits.

This will be the default behavior once go.mod is set to 1.24

maraino:
### Description

This commit fixes templates with policy identifiers on a program with the go.mod set to 1.24. With Go 1.24, the policies extension is marshaled using the values in the Policies field, but with Go 1.23 or a previous version, the marshaled field is PolicyIdentifiers.

This commit sets the same values on Policies and PolicyIdentifiers, ensuring that the templates with policies work in different versions of Go.

Fixes #738

### DIFF
--- a/x509util/marshal_utils.go
+++ b/x509util/marshal_utils.go
@@ -1,6 +1,7 @@
 package x509util
 
 import (
+	"crypto/x509"
 	"encoding/asn1"
 	"encoding/json"
 	"net"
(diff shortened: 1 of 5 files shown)


x509util/marshal_utils.go
```diff
@@ -129,6 +130,44 @@ func (m *MultiURL) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// MultiOID is a type used to unmarshal a JSON string or an array
+// of strings into a []x509.OID.
+type MultiOID []x509.OID
+
+// MarshalJSON implements the json.Marshaler interface for MultiOID.
+func (m MultiOID) MarshalJSON() ([]byte, error) {
```
hslatman, line 138: This and `UnmarshalJSON` may need a test

maraino: Fixed with 67ff3ee, I've also added tests with more complex policy templates.

hslatman (approved):

PR #770, opened 2025-05-17
Add support for 'DeleteKey' in CAPIKMS.

joshdrake:
💔Thank you!

### DIFF
--- a/kms/capi/ncrypt_windows.go
+++ b/kms/capi/ncrypt_windows.go
@@ -34,6 +34,7 @@ const (
 
 	// Key Storage Flags
 	NCRYPT_MACHINE_KEY_FLAG = 0x00000001
+	NCRYPT_SILENT_FLAG      = 0x00000040
 
 	// Errors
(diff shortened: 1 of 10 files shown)


kms/capi/ncrypt_windows.go
```diff
@@ -289,6 +291,18 @@ func nCryptOpenKey(provisionerHandle uintptr, containerName string, legacyKeySpe
...
 }
 
+func nCryptDeleteKey(keyHandle uintptr) error {
+	r, _, _ := procNCryptDeleteKey.Call(
+		keyHandle,
+		0,
+	)
+	if r != 0 {
+		return fmt.Errorf("NCryptDeleteKey returned %v", errNoToStr(uint32(r)))
+	}
```
hslatman, line 301: This might need a case similar to this before checking `r`:

```golang
if !errors.Is(err, windows.Errno(0)) {
    return nil, fmt.Errorf("NCryptDeleteKey returned %w", err)
}
```

hslatman (approved):

PR #781, opened 2025-06-03
Support storing certificates using mackms on data protection keychain.

joshdrake:
💔Thank you!

### DIFF
--- a/kms/mackms/mackms.go
+++ b/kms/mackms/mackms.go
@@ -83,8 +83,9 @@ type keySearchAttributes struct {
 }
 
 type certAttributes struct {
-	label        string
-	serialNumber *big.Int
+	label            string
+	useSecureEnclave bool
+	serialNumber     *big.Int
 }
 
 type algorithmAttributes struct {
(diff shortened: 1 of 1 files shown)


kms/mackms/mackms.go
```diff
@@ -964,25 +968,41 @@ func storeCertificate(label string, cert *x509.Certificate) error {
...
+		security.KSecAttrAccessControl: access,
+	}
+
+	// FIXME(josh): 'SecureEnclave' is the wrong terminology here,
+	// we should change this to something indicating that the data
+	// protection keychain should be used
+	if u.useSecureEnclave {
+		query[security.KSecUseDataProtectionKeychain] = cf.True
+	}
+
```
farhan687, line 988: Can we add `KSecAttrAccessControl` only when `se=true`? cert is not visible on safari when using `KSecAttrAccessControl`

```suggestion
	query := cf.Dictionary{
		security.KSecClass:    security.KSecClassCertificate,
		security.KSecValueRef: certRef,
	}

	// FIXME(josh): 'SecureEnclave' is the wrong terminology here,
	// we should change this to something indicating that the data
	// protection keychain should be used
	if u.useSecureEnclave {
		access, err := security.SecAccessControlCreateWithFlags(
			security.KSecAttrAccessibleAfterFirstUnlock,
			0,
		)
		if err != nil {
			return err
		}

		query[security.KSecAttrAccessControl] = access
		query[security.KSecUseDataProtectionKeychain] = cf.True
	}
```

PR #782, opened 2025-06-05
Remove superfluous error return values from all `randutil` functions

marten-seemann:
`crypto/rand.Reader` is **guaranteed** to never return an error, and to always read the entire slice. Hence, we don't need to check for errors.

From the [documentation](https://pkg.go.dev/crypto/rand#Read):
> Read fills b with cryptographically secure random bytes. It never returns an error, and always fills b entirely. 

See https://github.com/golang/go/issues/66821 for the related discussion in the Go standard library.

### DIFF
--- a/kms/sshagentkms/sshagentkms_test.go
+++ b/kms/sshagentkms/sshagentkms_test.go
@@ -294,7 +294,7 @@ func TestSSHAgentKMS_CreateSigner(t *testing.T) {
 		t.Fatal(err)
 	}
 	block, _ := pem.Decode(b)
-	block.Bytes, err = x509.DecryptPEMBlock(block, []byte("pass")) // nolint
+	block.Bytes, err = x509.DecryptPEMBlock(block, []byte("pass")) //nolint
 	if err != nil {
 		t.Fatal(err)
 	}


marten-seemann:
> Based on the release notes, the guarantee exists since 1.24.

As Filippo described in the Go issue I linked, on most platforms the underlying sources of randomness already have the property that they never return an error, with the notable exception of wasm. In that case, Go 1.24 just throws an (irrevocable?) panic. If we want to mimic that behavior on older Go versions, we could just panic here.

This would allow us to remove the error return value right now. I added that logic in 9fa5dae.

> There are also several external importers of the `randutil` package. We don't explicitly say `go.step.sm/crypto` is safe to use in your own projects, but there's also no disclaimer that people shouldn't use it (like we do for `cli-utils`), so it's fair that they're using it in their own projects. IMO we can make this "breaking" change in the same vein as the discussion in the Go repo: it'll will simplify its usage. Besides that, it'll also be more like Go itself.

We're using semver, and we're at `v0.x` anyway, so users have to expect breaking changes when the minor version number increases. Of course, I have no doubt that this will be a slightly annoying change to make, given how extensively we use this package, but then it's also a really, really easy fix.

hslatman:
> > Based on the release notes, the guarantee exists since 1.24.
> 
> As Filippo described in the Go issue I linked, on most platforms the underlying sources of randomness already have the property that they never return an error, with the notable exception of wasm. In that case, Go 1.24 just throws an (irrevocable?) panic. If we want to mimic that behavior on older Go versions, we could just panic here.

That is true for the `rand.Reader` inititialized by `rand` itself, but it's a global variable, and thus it is possible to override it. So there's no hard guarantee that it won't happen when a custom random generators is set at runtime. For example, we have have one backed by a TPM: https://github.com/smallstep/crypto/blob/master/tpm/random.go#L58. Afaik we don't use it as a replacement for `rand.Reader`, but it's also just a single example of why the assumption that it'll never return an error isn't always true. 

> This would allow us to remove the error return value right now. I added that logic in [9fa5dae](https://github.com/smallstep/crypto/commit/9fa5daed8f0df59fcb1d047bb4b0a89a2b0d9c8b).

Except that panic _is_recoverable. But overall I can agree with the approach.

> > There are also several external importers of the `randutil` package. We don't explicitly say `go.step.sm/crypto` is safe to use in your own projects, but there's also no disclaimer that people shouldn't use it (like we do for `cli-utils`), so it's fair that they're using it in their own projects. IMO we can make this "breaking" change in the same vein as the discussion in the Go repo: it'll will simplify its usage. Besides that, it'll also be more like Go itself.
> 
> We're using semver, and we're at `v0.x` anyway, so users have to expect breaking changes when the minor version number increases. Of course, I have no doubt that this will be a slightly annoying change to make, given how extensively we use this package, but then it's also a really, really easy fix.

Yeah, should generally be easy to fix, and I'll probably do that in the regular Dependabot review flow.

hslatman:
> > That is true for the `rand.Reader` inititialized by `rand` itself, but it's a global variable, and thus it is possible to override it. We do that in several projects to return fake/static random data.
> 
> It's one thing to do this for testing, I would argue that this is a really, really, really bad anti-pattern if used in production. You should not overwrite anything exposed by the standard library, ever (I've seen someone set `io.EOF = nil` to demonstrate the principle... just don't!). If you want a custom random data source, for a certain operation, use the respective `io.Reader` directly. For example, `tls.Config.Rand` allows you to do just that.

Sure, the fake/static random data is for testing, and where it's overridden, it generally is also reset at the end of the test.

Yet people do this with `http.DefaultTransport` and similar things. What I'm saying is that it would've been better if there was a guard that didn't allow this to happen in the first place: don't make it a global variable. If people can misuse, they will.

> As far as I can tell, we never do it outside of tests (very relieved to see that...): https://github.com/search?q=org%3Asmallstep%20%22rand.Reader%20%3D%22&type=code. And in many of these cases, we do this to simulate `rand.Reader` errors, which we'll be able to remove from our test suites altogether once this PR is merged.
> 
> In either case, if someone decides to set a custom `rand.Reader` that errors, both Go 1.23 and Go 1.24 will now panic, so I'd claim that we have all cases covered. I'm not concerned about recoverable vs. irrecoverable panics, the important property is that we exit the normal control flow, which is given in both cases.

It only irrevocably panics when `rand.Read` is used. Not when doing `rand.Reader.Read`, which I believe is what most usages will be doing, as the `rand.Reader` is passed around.

> As far as I can tell, there's nothing in the way of landing this change at this point. Should we merge this PR now?

I'd like @maraino's opinion on this one too.

maraino:
@marten-seemann What about creating a `randutil/v2` similar to `math/rand/v2`?

marten-seemann:
> @marten-seemann What about creating a `randutil/v2` similar to `math/rand/v2`?

I thought about that. We'd probably want to find a different name, it feels wrong to create a v2 if we aren't even at v1.

That said, I usually prefer to pay the one-time cost of updating code over the continuous cost of maintaining two slightly different APIs. I realize this is a breaking change that's mildly annoying to deal with, but I can offer to make this is painless as possible for everyone by signing up to update **all** of our repos within 24-48h of merging this PR (I just need someone to hit the Approve button a bunch of times). I already converted a few repos (see linked PRs) to get a feeling for how big this change is in practice.

hslatman:
> > @marten-seemann What about creating a `randutil/v2` similar to `math/rand/v2`?
> 
> I thought about that. We'd probably want to find a different name, it feels wrong to create a v2 if we aren't even at v1.
> 
> That said, I usually prefer to pay the one-time cost of updating code over the continuous cost of maintaining two slightly different APIs. I realize this is a breaking change that's mildly annoying to deal with, but I can offer to make this is painless as possible for everyone by signing up to update **all** of our repos within 24-48h of merging this PR (I just need someone to hit the Approve button a bunch of times). I already converted a few repos (see linked PRs) to get a feeling for how big this change is in practice.

`for everyone`: don't forget the external importers ;-)

kms/sshagentkms/sshagentkms_test.go
```diff
@@ -294,7 +294,7 @@ func TestSSHAgentKMS_CreateSigner(t *testing.T) {
 		t.Fatal(err)
 	}
 	block, _ := pem.Decode(b)
-	block.Bytes, err = x509.DecryptPEMBlock(block, []byte("pass")) //nolint
+	block.Bytes, err = x509.DecryptPEMBlock(block, []byte("pass")) // nolint
 	if err != nil {
```
hslatman: Is this `nolint` active at all? `//nolint` is in fact how it's required; otherwise it won't work. But we also usually have to add the specific linter behind it. So I guess this file is currently excluded, likely because it's a test.

randutil/random.go
```diff
@@ -22,80 +19,89 @@ func init() {
...
 // Salt generates a new random salt of the given size.
-func Salt(size int) ([]byte, error) {
+func Salt(size int) []byte {
 	salt := make([]byte, size)
-	_, err := io.ReadFull(rand.Reader, salt)
-	if err != nil {
-		return nil, errors.Wrap(err, "error generating salt")
+	// crypto/rand.Reader always fill the entire slices.
+	// Since Go 1.24, it is guaranteed to never return an error, see https://github.com/golang/go/issues/66821.
+	// This error check can be removed once we drop support for Go 1.23.
```
hslatman:
```suggestion
	// crypto/rand.Reader always fills the entire slice.
	// Since Go 1.24, it is guaranteed to never return an error, see https://github.com/golang/go/issues/66821.
	// This error check can be removed once we drop support for Go 1.23.
```

randutil/random.go
```diff
@@ -22,80 +19,89 @@ func init() {
...
 // Bytes generates a new byte slice of the given size.
-func Bytes(size int) ([]byte, error) {
+func Bytes(size int) []byte {
 	bytes := make([]byte, size)
-	_, err := io.ReadFull(rand.Reader, bytes)
-	if err != nil {
-		return nil, errors.Wrap(err, "error generating bytes")
+	// crypto/rand.Reader always fill the entire slices.
+	// Since Go 1.24, it is guaranteed to never return an error, see https://github.com/golang/go/issues/66821.
+	// This error check can be removed once we drop support for Go 1.23.
```
hslatman:
```suggestion
	// crypto/rand.Reader always fills the entire slice.
	// Since Go 1.24, it is guaranteed to never return an error, see https://github.com/golang/go/issues/66821.
	// This error check can be removed once we drop support for Go 1.23.
```

randutil/random.go
```diff
@@ -22,80 +19,89 @@ func init() {
...
-func UUIDv4() (string, error) {
+func UUIDv4() string {
 	var uuid [16]byte
-	_, err := io.ReadFull(rand.Reader, uuid[:])
-	if err != nil {
-		return "", errors.Wrap(err, "error generating uuid")
+	// crypto/rand.Reader is guaranteed to never return an error, and always fill the entire slices
+	// crypto/rand.Reader always fill the entire slices.
+	// Since Go 1.24, it is guaranteed to never return an error, see https://github.com/golang/go/issues/66821.
+	// This error check can be removed once we drop support for Go 1.23.
```
hslatman:
```suggestion
	// crypto/rand.Reader is guaranteed to never return an error, and always fill the entire slices
	// crypto/rand.Reader always fills the entire slice.
	// Since Go 1.24, it is guaranteed to never return an error, see https://github.com/golang/go/issues/66821.
	// This error check can be removed once we drop support for Go 1.23.
```

kms/sshagentkms/sshagentkms_test.go
marten-seemann: My IDE must have done this automatically, I didn't even notice. `x509.DecryptPEMBlock` is deprecated as it's insecure, but this is in tests, so 🤷‍♂️

```suggestion
	block.Bytes, err = x509.DecryptPEMBlock(block, []byte("pass"))
	if err != nil {
```

randutil/random.go
marten-seemann: GitHub broke the "Commit suggestion" UI, so I fixed it manually in 17c0809136a9c61cf660b3f7fc2640586a23afdd.

kms/sshagentkms/sshagentkms_test.go
marten-seemann: Removed it in 6beadb23ed11ededcfe34062db232014921d6fe7.

hslatman (approved):
I'll approve. Let's await @maraino's verdict too.

hslatman:
> This is covered by Go's backwards compatibility guarantee. They will not be able to ever ship a `crypto/rand.Reader` that exhibits a different behavior.

I find this a pretty strong statement. At the code level, the interface still says it can return an error. I'm certain the Go team will critically evaluate changes in underlying behavior, but I don't believe there's a formal guarantee. Even today, on WASIP, it can return errors if the underlying `random_get` returns an error; there doesn't seem to be an additional guard for that case.

> This is impossible with the `io.Reader` interface, which seems like a non-starter for an object that reads bytes into a slice.

Thus the suggestion to have a different package, with a different interface, API, or way of getting access to the random reader. 

> My goal was to reduce complexity by not testing for errors that are guaranteed to never occur (and with that, remove a whole bunch of test cases that simulate an error-ing `rand.Reader`). I still think that we're needlessly complicating our code base, but in the end, I don't care enough to push on this any further.

In places where we control the `rand.Reader`, thus most of our stack, actually, we could still benefit by using the (to be implemented) panicking functions.

PR #793, opened 2025-06-18
Add support for the Data Protection Keychain

maraino:
### Description

This PR adds support for storing certificates in the Data Protection Keychain. This functionality is only available for code-signed applications.

Additionally, this commit includes a fix that enables retrieving a key from the Data Protection Keychain or not based on URI parameters. 

This last fix slightly changes the behavior: previously, keys stored in the Secure Enclave could be retrieved without specifying `se=true`, but it was not possible to retrieve the proper key if a key in the Secure Enclave shares the same label as a key in the login keychain.

### DIFF
--- a/internal/darwin/security/security_darwin.go
+++ b/internal/darwin/security/security_darwin.go
@@ -394,6 +394,7 @@ func goOSStatus(status C.OSStatus) error {
 		message = goString(C.CFStringRef(ref.Value))
 		defer ref.Release()
 	}
+	fmt.Println(message, int(status))
 	return osStatusError{
 		code:    int(status),
 		message: message,
--- a/internal/darwin/security/security_darwin.go
+++ b/internal/darwin/security/security_darwin.go
@@ -394,7 +394,6 @@ func goOSStatus(status C.OSStatus) error {
 		message = goString(C.CFStringRef(ref.Value))
 		defer ref.Release()
 	}
-	fmt.Println(message, int(status))
 	return osStatusError{
 		code:    int(status),
 		message: message,
--- a/kms/mackms/mackms.go
+++ b/kms/mackms/mackms.go
@@ -625,7 +625,8 @@ func (k *MacKMS) DeleteCertificate(req *apiv1.DeleteCertificateRequest) error {
 	}
 
 	query := cf.Dictionary{
-		security.KSecClass: security.KSecClassCertificate,
+		security.KSecClass:      security.KSecClassCertificate,
+		security.KSecMatchLimit: security.KSecMatchLimitOne,
 	}
 	if u.label != "" {
 		cfLabel, err := cf.NewString(u.label)
--- a/kms/mackms/mackms.go
+++ b/kms/mackms/mackms.go
@@ -50,11 +50,6 @@ const Scheme = string(apiv1.MacKMS)
 // the keys.
 var DefaultTag = "com.smallstep.crypto"
 
-// UseDataProtectionKeychain defines if mackms will use the Data Protection
-// keychain by default. This option requires the application to be code-signed
-// with the appropriate entitlements.
-var UseDataProtectionKeychain = false
-
 type keyAttributes struct {
 	label            string
 	tag              string
@@ -146,11 +141,29 @@ var signatureAlgorithmMapping = map[apiv1.SignatureAlgorithm]algorithmAttributes
 //     DER representation of an RSA public key using the PKCS #1 format or the
 //     SHA-1 of the uncompressed ECDSA point according to SEC 1, Version 2.0,
 //     Section 2.3.4.
-type MacKMS struct{}
+type MacKMS struct {
+	useDataProtectionKeychain bool
+}
 
-// New returns a new SoftKMS.
-func New(context.Context, apiv1.Options) (*MacKMS, error) {
-	return &MacKMS{}, nil
+// New returns a new [MacKMS] instance. In code-signed applications, it is
+// possible to initialize [MacKMS] with a URI that specifies whether to use the
+// Data Protection Keychain for loading and storing certificates by default.
+//
+// The [New] function accepts the following URI format:
+//   - mackms:keychain=dataProtection
+func New(_ context.Context, opts apiv1.Options) (*MacKMS, error) {
+	var useDataProtectionKeychain bool
+	if opts.URI != "" {
+		u, err := uri.ParseWithScheme(Scheme, opts.URI)
+		if err != nil {
+			return nil, err
+		}
+		useDataProtectionKeychain = isDataProtectionKeychain(u.Get("keychain"), false)
+	}
+
+	return &MacKMS{
+		useDataProtectionKeychain: useDataProtectionKeychain,
+	}, nil
 }
 
 func init() {
@@ -381,7 +394,7 @@ func (k *MacKMS) LoadCertificate(req *apiv1.LoadCertificateRequest) (*x509.Certi
 	}
 
 	// Require label or serial
-	u, err := parseCertURI(req.Name, true)
+	u, err := parseCertURI(req.Name, k.useDataProtectionKeychain, true)
 	if err != nil {
 		return nil, fmt.Errorf("mackms LoadCertificate failed: %w", err)
 	}
@@ -421,7 +434,7 @@ func (k *MacKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
 
 	// Do not require any parameter. Using mackms: is allowed as macOS will set
 	// the commonName as label.
-	u, err := parseCertURI(req.Name, false)
+	u, err := parseCertURI(req.Name, k.useDataProtectionKeychain, false)
 	if err != nil {
 		return fmt.Errorf("mackms StoreCertificate failed: %w", err)
 	}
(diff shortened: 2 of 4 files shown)


internal/darwin/security/security_darwin.go
```diff
@@ -394,6 +394,7 @@ func goOSStatus(status C.OSStatus) error {
 		message = goString(C.CFStringRef(ref.Value))
 		defer ref.Release()
 	}
+	fmt.Println(message, int(status))
```
hslatman: Don't forget to remove

kms/mackms/mackms.go
```diff
@@ -48,6 +50,11 @@ const Scheme = string(apiv1.MacKMS)
 // the keys.
 var DefaultTag = "com.smallstep.crypto"
 
+// UseDataProtectionKeychain defines if mackms will use the Data Protection
+// keychain by default. This option requires the application to be code-signed
+// with the appropriate entitlements.
+var UseDataProtectionKeychain = false
```
hslatman: Instead of setting this on the package level, can it be provided when instantiating an instance instead? That's closer to how we do things for e.g. the TPM KMS with CAPI feedthrough. Or is there a specific reason this needs to be defined as a package global variable?

kms/mackms/mackms.go
```diff
@@ -427,9 +464,9 @@ func (k *MacKMS) LoadCertificateChain(req *apiv1.LoadCertificateChainRequest) ([
 		return nil, fmt.Errorf("mackms LoadCertificateChain failed: %w", err)
 	}
 
-	cert, err := loadCertificate(u.label, u.serialNumber, nil)
+	cert, err := loadCertificate(u, nil)
 	if err != nil {
-		return nil, fmt.Errorf("mackms LoadCertificateChain failed: %w", apiv1Error(err))
+		return nil, fmt.Errorf("mackms LoadCertificateChain failed1: %w", apiv1Error(err))
```
hslatman:
```suggestion
		return nil, fmt.Errorf("mackms LoadCertificateChain failed: %w", apiv1Error(err))
```

kms/mackms/mackms.go
```diff
@@ -479,15 +517,27 @@ func (k *MacKMS) StoreCertificateChain(req *apiv1.StoreCertificateChainRequest)
...
+				u.useDataProtectionKeychain = true
+			}
+		}
+	}
+
 	// Store the certificate and update the label if required
-	if err := storeCertificate(u.label, req.CertificateChain[0]); err != nil {
-		return fmt.Errorf("mackms StoreCertificateChain failed: %w", apiv1Error(err))
+	if err := storeCertificate(u, req.CertificateChain[0]); err != nil {
+		return fmt.Errorf("mackms StoreCertificateChain failed1: %w", apiv1Error(err))
```
hslatman:
```suggestion
		return fmt.Errorf("mackms StoreCertificateChain failed: %w", apiv1Error(err))
```

kms/mackms/mackms.go
```diff
@@ -479,15 +517,27 @@ func (k *MacKMS) StoreCertificateChain(req *apiv1.StoreCertificateChainRequest)
...
 	}
 
 	// Store the rest of the chain but do not fail if already exists
+	chainURI := &certAttributes{useDataProtectionKeychain: u.useDataProtectionKeychain}
 	for _, cert := range req.CertificateChain[1:] {
-		if err := storeCertificate("", cert); err != nil && !errors.Is(err, security.ErrAlreadyExists) {
-			return fmt.Errorf("mackms StoreCertificateChain failed: %w", err)
+		if err := storeCertificate(chainURI, cert); err != nil && !errors.Is(err, security.ErrAlreadyExists) {
+			return fmt.Errorf("mackms StoreCertificateChain failed2: %w", err)
```
hslatman:
```suggestion
			return fmt.Errorf("mackms StoreCertificateChain failed: %w", err)
```

kms/mackms/mackms.go
```diff
@@ -966,10 +1075,29 @@ func storeCertificate(label string, cert *x509.Certificate) error {
...
 		security.KSecClass:    security.KSecClassCertificate,
 		security.KSecValueRef: certRef,
-	})
+	}
+	// Apple recommends to set KSecUseDataProtectionKeychain parameters always
+	// to true for all keychains operations. This option requires to use a
+	// code-signed application.
+	if u.useDataProtectionKeychain {
+		dict[security.KSecUseDataProtectionKeychain] = cf.True
+		// The data in the keychain item cannot be accessed after a restart
```
hslatman: Is it possible to detect the required entitlements are available in the program? If so, when the intention is to use the data protection keychain, and the application isn't codesigned / missing entitlements, can it return an error indicating that case? This may not be necessary if errors from the OS are clear enough, but I've seen enough OSError codes that weren't overly clear 😅

kms/mackms/mackms.go
```diff
@@ -357,6 +367,14 @@ func (k *MacKMS) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, er
 //   - mackms:label=test@example.com
 //   - mackms:serial=2c273934eda8454d2595a94497e2395a
 //   - mackms:label=test@example.com;serial=2c273934eda8454d2595a94497e2395a
+//
+// On code-signed applications, it is possible to use the Data Protection
+// Keychain by default if [UseDataProtectionKeychain] is set to true, you can
+// also select the keychain using the "keychain" attribute:
```
hslatman:
```suggestion
// On code-signed applications, it is possible to use the Data Protection
// Keychain by default if [UseDataProtectionKeychain] is set to true, you can
// also select the keychain using the "keychain" attribute:
```
```suggestion
// On code-signed applications, it is possible to use the Data Protection
// Keychain by default if [UseDataProtectionKeychain] is set to true. You can
// also select the keychain using the "keychain" attribute:
```

kms/mackms/mackms.go
```diff
@@ -386,6 +404,14 @@ func (k *MacKMS) LoadCertificate(req *apiv1.LoadCertificateRequest) (*x509.Certi
 //   - "mackms:" will use the common name
 //   - "mackms:label=my-label" will use "my-label"
 //   - "mackms:my-label" will use the "my-label"
+//
+// On code-signed applications, it is possible to use the Data Protection
+// Keychain by default if [UseDataProtectionKeychain] is set to true, you can
+// also select the keychain using the "keychain" attribute:
```
hslatman:
```suggestion
// On code-signed applications, it is possible to use the Data Protection
// Keychain by default if [UseDataProtectionKeychain] is set to true, you can
// also select the keychain using the "keychain" attribute:
```
```suggestion
// On code-signed applications, it is possible to use the Data Protection
// Keychain by default if [UseDataProtectionKeychain] is set to true. You can
// also select the keychain using the "keychain" attribute:
```

kms/mackms/mackms.go
```diff
@@ -924,6 +1026,13 @@ func loadCertificate(label string, serialNumber *big.Int, subjectKeyID []byte) (
 		defer cfSubjectKeyID.Release()
 		query[security.KSecAttrSubjectKeyID] = cfSubjectKeyID
 	}
+	// Apple recommends to set this parameters always to true for all keychains
+	// operations. This requires a code-signed application.
```
hslatman:
```suggestion
	// Apple recommends to set this parameters always to true for all keychains
	// operations. This requires a code-signed application.
```
```suggestion
	// Apple recommends to set this parameters always to true for all keychain
	// operations. This requires a code-signed application.
```

kms/mackms/mackms.go
```diff
@@ -966,10 +1075,29 @@ func storeCertificate(label string, cert *x509.Certificate) error {
...
 	// Adding the label here doesn't have any effect. Apple Keychain always uses
 	// the commonName.
-	attributes, err := cf.NewDictionary(cf.Dictionary{
+	dict := cf.Dictionary{
 		security.KSecClass:    security.KSecClassCertificate,
 		security.KSecValueRef: certRef,
-	})
+	}
+	// Apple recommends to set KSecUseDataProtectionKeychain parameters always
+	// to true for all keychains operations. This option requires to use a
```
hslatman:
```suggestion
	// to true for all keychain operations. This option requires to use a
```

kms/mackms/mackms.go
joshdrake: Additionally, what benefit is there to allow specifying the default behavior like this? Shouldn't we always prefer to derive `UseDataProtectionKeychain` based on where the key resides? There should never be any reason to store the certificate in a file-based keychain when the key is on the data-protection keychain, and vice-versa.

kms/mackms/mackms.go
maraino: Apparently it should be possible to know this using [SecCodeCheckValidity ](https://developer.apple.com/documentation/security/seccodecheckvalidity(_:_:_:)?language=objc). We can add this in a different PR. 

The current behavior is:
- When loading a certificate you will get a not found error. Something like:
```
mackms LoadCertificate failed: not found
```
- When storing a certificate you will get an error showing that an entitlement is missing:
```
mackms StoreCertificateChain failed: OSStatus -34018: A required entitlement isn't present.
```

kms/mackms/mackms.go
maraino: @hslatman: Fixed with caf7865. 

> Shouldn't we always prefer to derive UseDataProtectionKeychain based on where the key resides?

@joshdrake: We are doing that, if you don't pass any keychain to `StoreCertificate` or `StoreCertificateChain` it will look for the key and store it in the proper location.

internal/darwin/security/security_darwin.go
maraino: Fixed with https://github.com/smallstep/crypto/commit/caf78652d0f2d7250c90ee533a6f8208437251a9.

joshdrake (dismissed):

joshdrake (approved):

PR #802, opened 2025-07-11
Correctly set the machine store flag when requested

areed:
Fix the NCRYPT_MACHINE_KEY_FLAG constant and use it in calls for key operations when the machine store is specified in the uri. Validate the store arg is compatible with the provider.

```
step-kms-plugin create capi:store-location=machine;key=a1 --kty EC --crv P-521
Error: failed to create key: unable to set key NCRYPT_KEY_TYPE_PROPERTY: NCryptSetProperty Key Type invalid value type int
```

#### Why is this important to the project (if not answered above):

Creating a key in the machine store fails. Getting or deleting a key from the machine store silently gets or deletes the key of the same name in the user store.

Windows

💔Thank you!


kms/capi/ncrypt_windows.go
```diff
@@ -33,7 +33,7 @@ const (
 	NCRYPT_PCP_USAGE_AUTH_PROPERTY  = "PCP_USAGEAUTH"
 
 	// Key Storage Flags
-	NCRYPT_MACHINE_KEY_FLAG = 0x00000001
+	NCRYPT_MACHINE_KEY_FLAG = 0x00000020
```
areed, line 36: https://learn.microsoft.com/en-us/windows/win32/seccng/key-storage-property-identifiers#ncrypt_key_type_property

kms/capi/capi.go
```diff
@@ -46,6 +46,17 @@ const (
 	SkipFindCertificateKey = "skip-find-certificate-key" // skips looking up certificate private key when storing a certificate
 )
 
+const (
+	SoftwareProvider  = "Microsoft Software Key Storage Provider"
+	SmartCardProvider = "Microsoft Smart Card Key Storage Provider"
+	PlatformProvider  = "Microsoft Platform Crypto Provider"
```
maraino: ncrypt_windows.go provides the same constants with a different name, perhaps we want to use those instead of creating new ones. We should also use a constant on New() to set the default provider.

maraino (changes requested):

kms/capi/capi.go
areed: 2fe35c1

maraino (approved):

PR #803, opened 2025-07-12
Fix conversion of raw ecdsa p-521 signature to asn encoded

areed:
Signature verification fails on Windows with capi kms when signing with EC p-521 keys since r and s are supposed to be 66 bytes each.

Fix https://github.com/smallstep/cli/issues/1453.

💔Thank you!


maraino (approved):
lgtm

PR #811, opened 2025-08-03
Accept nonRepudiation when unmarshalling keyUsage

neverpanic:
#### Name of feature:
Accepting `NonRepudiation` when unmarshalling `keyUsage`.

Better user experience for users that are used to the old name (e.g., from OpenSSL, which still uses the old name exclusively).

#### Is there documentation on how to use this feature? If so, where?
I didn't add any, because I don't think this should be advertised — but it's nice when it works when somebody attempts it.

smallstep/certificates#2348

#### Description
ITU-T X.509 (10/2019) section 9.2.2.3 [1] defines 'contentCommitment' as the current name for what had previously been called 'nonRepudiation', and deprecates the old name:

> It is not incorrect to refer to this keyUsage bit using the identifier
> nonRepudiation. However, the use of this identifier has been
> deprecated.

The old name is still widely used, though — for example, OpenSSL still supports nonRepudiation exclusively. It is therefore not surprising that users attempt to specify that and just get an error message in return. See for example 
Modify the JSON unmarshalling to accept the old name, and transparently convert it to the new one. Keep marshalling the same, since we want to be conservative in what we send, but liberal in what we accept.

[1]: https://www.itu.int/rec/T-REC-X.509-201910-I/en

Related: smallstep/certificates#2348

### DIFF
--- a/x509util/extensions.go
+++ b/x509util/extensions.go
@@ -40,6 +40,8 @@ const (
 	KeyUsageCRLSign           = "crlSign"
 	KeyUsageEncipherOnly      = "encipherOnly"
 	KeyUsageDecipherOnly      = "decipherOnly"
+
+	keyUsageNonRepudiation = "nonRepudiation" // old name for contentCommitment
 )
 
 // Names used for extended key usages.
@@ -727,7 +729,9 @@ func (k *KeyUsage) UnmarshalJSON(data []byte) error {
 		switch convertName(s) {
 		case convertName(KeyUsageDigitalSignature):
 			ku = x509.KeyUsageDigitalSignature
-		case convertName(KeyUsageContentCommitment):
+		// support legacy name for better user experience for users coming from
+		// OpenSSL
+		case convertName(KeyUsageContentCommitment), convertName(keyUsageNonRepudiation):
 			ku = x509.KeyUsageContentCommitment
 		case convertName(KeyUsageKeyEncipherment):
 			ku = x509.KeyUsageKeyEncipherment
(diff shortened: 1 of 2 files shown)


x509util/extensions.go
```diff
@@ -727,6 +727,10 @@ func (k *KeyUsage) UnmarshalJSON(data []byte) error {
 		switch convertName(s) {
 		case convertName(KeyUsageDigitalSignature):
 			ku = x509.KeyUsageDigitalSignature
+		case convertName("nonRepudiation"):
+			// support legacy name for better user experience for users coming
+			// from OpenSSL
+			fallthrough
 		case convertName(KeyUsageContentCommitment):
```
hslatman: Could you combine this into the single case statement instead of the `fallthrough`? I don't expect the code lines to change order, but just in case. Can you also make the `"nonRepudiation"` a non-exported constant? I.e. `keyUsageNonRepudiation = "nonRepudiation" // old name for contentCommitment key usage`

hslatman: Apparently we even enforce that style: https://github.com/smallstep/crypto/actions/runs/16708934820/job/47321789675?pr=811#step:9:32 😅

neverpanic: Should be fixed now.

neverpanic:
> @neverpanic just a minor linting issue left now

Should be fixed now — would be nice if `golangci-lint` would actually print the suggested formatting changes, rather than just saying "it's wrong"…

PR #873, opened 2025-10-14
Add support for additional CAPI certificate context properties

joshdrake:
This PR adds functionality to the `capi` KMS to support setting the "friendly name" and "description" certificate properties. The `tpmkms` passes these through as needed.

💔Thank you!

### DIFF
--- a/kms/apiv1/options.go
+++ b/kms/apiv1/options.go
@@ -84,7 +84,7 @@ type CertificateDeleter interface {
 // Notice: This API is EXPERIMENTAL and may be changed or removed in a later
 // release.
 type CleaningCertificateManager interface {
-	Cleanup(issuer, storeLocation, store string, subjectRaw []byte) error
+	Cleanup(req *CleanupCertificatesRequest) error
 }
 
 // NameValidator is an interface that KeyManager can implement to validate a
--- a/kms/apiv1/options.go
+++ b/kms/apiv1/options.go
@@ -76,6 +76,17 @@ type CertificateDeleter interface {
 	DeleteCertificate(req *DeleteCertificateRequest) error
 }
 
+// CleaningCertificateManager is an optional interface for KMS implementations
+// that support cleaning up expired certificates from a certificate store.
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
+// release.
+type CleaningCertificateManager interface {
+	Cleanup(issuer, storeLocation, store string, subjectRaw []byte) error
+}
+
 // NameValidator is an interface that KeyManager can implement to validate a
 // given name or URI.
 type NameValidator interface {
--- a/kms/capi/capi.go
+++ b/kms/capi/capi.go
@@ -433,11 +433,9 @@ func (k *CAPIKMS) getCertContext(u *uriAttributes) (*windows.CertContext, error)
 		}
 	case len(u.keyID) > 0:
 		if handle, err = findCertificateBySubjectKeyID(st, u.keyID); err != nil {
-			return nil, err
-		}
-
-		if handle == nil && !canLookupByIssuer {
-			return nil, apiv1.NotFoundError{Message: fmt.Sprintf("certificate with %s=%s not found", KeyIDArg, u.keyID)}
+			if !errors.Is(err, apiv1.NotFoundError{}) || !canLookupByIssuer {
+				return nil, err
+			}
 		}
 	case u.containerName != "":
 		key, err := k.GetPublicKey(&apiv1.GetPublicKeyRequest{
@@ -453,11 +451,9 @@ func (k *CAPIKMS) getCertContext(u *uriAttributes) (*windows.CertContext, error)
 			return nil, fmt.Errorf("error generating SubjectKeyID: %w", err)
 		}
 		if handle, err = findCertificateBySubjectKeyID(st, keyID); err != nil {
-			return nil, err
-		}
-
-		if handle == nil && !canLookupByIssuer {
-			return nil, apiv1.NotFoundError{Message: fmt.Sprintf("certificate with %s=%s not found", KeyIDArg, u.keyID)}
+			if !errors.Is(err, apiv1.NotFoundError{}) || !canLookupByIssuer {
+				return nil, err
+			}
 		}
 	}
 
--- a/kms/capi/capi.go
+++ b/kms/capi/capi.go
@@ -437,7 +437,7 @@ func (k *CAPIKMS) getCertContext(u *uriAttributes) (*windows.CertContext, error)
 			0,
 			findCertID,
 			uintptr(unsafe.Pointer(&searchData)), nil)
-		if err != nil && !canLookupByIssuer() {
+		if err != nil && !canLookupByIssuer {
 			return nil, fmt.Errorf("findCertificateInStore failed: %w", err)
 		}
 
@@ -863,12 +863,12 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
 		cryptFindCertificateKeyProvInfo(certContext)
 	}
 
-	if friendlyName := u.Get(FriendlyNameArg); friendlyName != "" {
-		cryptSetCertificateFriendlyName(certContext, friendlyName)
+	if u.friendlyName != "" {
+		cryptSetCertificateFriendlyName(certContext, u.friendlyName)
 	}
 
-	if description := u.Get(DescriptionArg); description != "" {
-		cryptSetCertificateDescription(certContext, description)
+	if u.description != "" {
+		cryptSetCertificateDescription(certContext, u.description)
 	}
 
 	st, err := windows.CertOpenStore(
@@ -906,6 +906,8 @@ func (k *CAPIKMS) StoreCertificateChain(req *apiv1.StoreCertificateChainRequest)
 			HashArg:                []string{fp},
 			StoreLocationArg:       []string{u.storeLocation},
 			StoreNameArg:           []string{u.storeName},
+			FriendlyNameArg:        []string{u.friendlyName},
+			DescriptionArg:         []string{u.description},
 			SkipFindCertificateKey: []string{strconv.FormatBool(u.skipFindCertificateKey)},
 		}).String(),
 		Certificate: leaf,
--- a/kms/capi/capi.go
+++ b/kms/capi/capi.go
@@ -404,6 +404,10 @@ func (k *CAPIKMS) getCertContext(u *uriAttributes) (*windows.CertContext, error)
 		return nil, fmt.Errorf("CertOpenStore for the %q store %q returned: %w", u.storeLocation, u.storeName, err)
 	}
 
+	// if issuer + any of the other fields in the list below is provided, then attempt a second certificate lookup when
+	// lookup by KeyID fails (not found). This fix an issue when looking up device certificates, as in that case the KeyID is
+	// derived from a randomly generate string each time agent runs, thus not being able to find certificates installed from
+	// a previous run.
 	canLookupByIssuer := u.issuerName != "" && (u.serialNumber != "" || u.subjectCN != "" || u.friendlyName != "" || u.description != "")
 	var handle *windows.CertContext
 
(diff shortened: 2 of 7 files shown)


kms/tpmkms/tpmkms.go
```diff
@@ -728,13 +731,24 @@ func (k *TPMKMS) loadCertificateChainFromWindowsCertificateStore(req *apiv1.Load
...
+	switch {
+	case subjectKeyID != "":
+		uv.Set("key-id", subjectKeyID)
+	case o.issuer != "":
+		uv.Set("issuer", o.issuer)
+	case o.friendlyName != "":
+		uv.Set("friendly-name", o.friendlyName)
+	case o.description != "":
+		uv.Set("description", o.description)
 	}
```
darkfronza: Are those mutually exclusive?

kms/capi/capi.go
```diff
@@ -349,6 +353,10 @@ func (k *CAPIKMS) getCertContext(req *apiv1.LoadCertificateRequest) (*windows.Ce
 		return nil, fmt.Errorf("CertOpenStore for the %q store %q returned: %w", storeLocation, storeName, err)
 	}
 
+	canLookupByIssuer := func() bool {
+		return issuerName != "" && (serialNumber != "" || subjectCN != "" || friendlyName != "" || description != "")
+	}
```
joshdrake: Why a function here, instead of eg: a `const`?

kms/tpmkms/tpmkms.go
joshdrake: It might not make much sense, but if we're going off the underlying CAPI, they shouldn't be. In our calling code paths, we may want make to only ever provide one, so that we can ensure our desired property is used for retrieval (eg: the `description`), without worrying about the underlying implementation in `capikms`.

joshdrake (commented):
One nit/question, otherwise lgtm. Need another reviewer to be able to merge, cc @hslatman

kms/tpmkms/tpmkms.go
darkfronza: I see, makes sense.

kms/capi/capi.go
darkfronza: Ah, this was relic from previous code when I was updating it, fixed here https://github.com/smallstep/crypto/pull/873/commits/b78b8bba5796f4b55299c14439f86b3b1901c62d

kms/capi/capi.go
```diff
@@ -426,13 +434,20 @@ func (k *CAPIKMS) getCertContext(u *uriAttributes) (*windows.CertContext, error)
 			0,
 			findCertID,
 			uintptr(unsafe.Pointer(&searchData)), nil)
-		if err != nil {
+		if err != nil && !canLookupByIssuer {
```
smst-jeff: I think in the case where `err` != nil, `canLookupByIssuer` == true, and `handle` == nil, this err is dropped.

If that is expected, I think it should be documented. Tracing the path of `err` here is a bit tedious.

kms/capi/capi.go
```diff
@@ -426,13 +434,20 @@ func (k *CAPIKMS) getCertContext(u *uriAttributes) (*windows.CertContext, error)
...
 			return nil, fmt.Errorf("findCertificateInStore failed: %w", err)
```
smst-jeff, line 475: Is this error message still correct? Not sure how `canLookupByIssuer` should be included.

kms/capi/ncrypt_windows.go
```diff
@@ -606,6 +611,102 @@ func cryptFindCertificateKeyContainerName(certContext *windows.CertContext) (str
 	return "", nil
 }
 
+func certSetCertificateContextProperty(certContext *windows.CertContext, propID uint32, pvData uintptr) error {
+	r0, _, err := procCertSetCertificateContextProperty.Call(
```
smst-jeff, line 642: What is `r0`? I think this could use a more descriptive name.

kms/capi/ncrypt_windows.go
```diff
@@ -606,6 +611,102 @@ func cryptFindCertificateKeyContainerName(certContext *windows.CertContext) (str
...
+		uintptr(unsafe.Pointer(certContext)),
+		uintptr(propID),
+		0,
+		pvData,
+	)
+
+	if r0 == 0 {
+		return err
+	}
+	return nil
```
smst-jeff, line 652: Is `err` assumed to be non-nil here?

If `err` != nil, and `r0` is != 0, is it ok to drop the error?

kms/capi/capi.go
darkfronza: I think I can improve this a little bit, the logic here is as follows so you have more context: if looking up a certificate by keyID fails, then if user also provided issuer+field (where field could be serial, description among others), then we can attempt another certificate lookup by those fields.
This was implemented to solve the issue with duplicate certs on windows storage, as KeyID for device certificate is derived from a random string each time agent runs, thus not being able to find certificates from a previous agent run.
I think it's better to adjust this and try the second lookup only if handle was not found, not if an error happened, adjusted here: https://github.com/smallstep/crypto/pull/873/commits/59b1bc251c4955c6279720758b8dc732bbb4f489

kms/capi/capi.go
darkfronza: Fixed https://github.com/smallstep/crypto/pull/873/commits/59b1bc251c4955c6279720758b8dc732bbb4f489
Explanation on the thread above.

kms/capi/ncrypt_windows.go
darkfronza: It's actually r1, this invokes a system call on windows (whose procedure is found in a DLL), the system call returns r1, r2, error, in general, where r1 represents the return value status from the procedure stored in a register (e.g. on Linux the %rax value), the semantic value of this depends on the procedure invoked, r2 is usually not used but kept for compatibility with platforms that return status on more registers, and the 3rd value return is actually an error, on windows the error is always non nil so we must check r0(r1).
I think we can leave this low level stuff as is, or perhaps considering refactor this in another pr.

kms/capi/ncrypt_windows.go
darkfronza: This invokes `CertSetCertificateContextProperty` which returns a bool, on success it returns true.
But since this is system call invocation the bool is cast to int, false is zero, so in that case we actually return the error.
We handle this as syscall on windows always return a non nil error even if the function succeeds

kms/capi/capi.go
smst-jeff: Appreciate the comment!

kms/capi/ncrypt_windows.go
smst-jeff: Ok, that all makes sense. Can you add a comment to these funcs explaining it? With the information you provided above, it's clear, but without it, it's not clear what's happening.

kms/capi/ncrypt_windows.go
smst-jeff: Similar to the above, I think comments helps a lot here.

kms/capi/ncrypt_windows.go
```diff
@@ -62,9 +62,11 @@ const (
 	compareShift            = 16                                              // CERT_COMPARE_SHIFT
 	compareSHA1Hash         = 1                                               // CERT_COMPARE_SHA1_HASH
 	compareCertID           = 16                                              // CERT_COMPARE_CERT_ID
+	compareProp             = 5                                               // CERT_COMPARE_CERT_ID
```
hslatman:
```suggestion
	compareProp             = 5                                               // CERT_COMPARE_PROPERTY
```

kms/capi/capi.go
```diff
@@ -392,11 +398,17 @@ func (k *CAPIKMS) getCertContext(u *uriAttributes) (*windows.CertContext, error)
...
+		uintptr(unsafe.Pointer(wide(u.storeName))),
+	)
 	if err != nil {
 		return nil, fmt.Errorf("CertOpenStore for the %q store %q returned: %w", u.storeLocation, u.storeName, err)
 	}
 
+	// if issuer + any of the other fields in the list below is provided, then attempt a second certificate lookup when
+	// lookup by KeyID fails (not found). This fix an issue when looking up device certificates, as in that case the KeyID is
+	// derived from a randomly generate string each time agent runs, thus not being able to find certificates installed from
+	// a previous run.
```
hslatman:
```suggestion
	// if issuer + any of the other fields in the list below is provided, then attempt a second certificate lookup when
	// lookup by KeyID fails (not found).
```

kms/tpmkms/tpmkms.go
```diff
@@ -950,6 +953,14 @@ func (k *TPMKMS) storeCertificateChainToWindowsCertificateStore(req *apiv1.Store
 	if o.store != "" {
 		store = o.store
 	}
+
+	if o.issuer != "" && (o.description == "device-id" || o.description == "step-agent-id") {
+		// best-effort: log failure but do not block the store operation
+		if err := k.cleanupExpiredCertificatesFromWindowsCertificateStore(o.issuer, location, store, req.CertificateChain[0].RawSubject); err != nil {
+			_ = err // TODO: replace with structured logging once a logger is available in this context
+		}
+	}
```
hslatman: This leaks application logic into library code. It would be better if there's an alternative way to go about this. 

Given that it's gated on this conditional, we could go ahead with the current implementation to meet time constraints, but we should refactor if possible.

hslatman (dismissed):
Approving it, with the remark that the application logic ideally would be moved out of here.

kms/tpmkms/tpmkms.go
```diff
@@ -975,6 +981,54 @@ func (k *TPMKMS) storeCertificateChainToWindowsCertificateStore(req *apiv1.Store
 	})
 }
 
+// Cleanup implements [apiv1.CleaningCertificateManager]. It finds all certificates
+// in the Windows certificate store issued to the given subject by the given issuer,
+// and deletes any that have already expired.
+func (k *TPMKMS) Cleanup(issuer, storeLocation, store string, subjectRaw []byte) error {
+	if !k.usesWindowsCertificateStore() {
+		return apiv1.NotImplementedError{}
```
hslatman: In this case, I think returning early with `nil` (with a comment explaining why) would be OK (for now). 

We've used the pattern to ascertain a KMS implements a certain interface before calling into a function. Doing that now would result in an error when the KMS happens to not be configured with a backing certificate store. Unless we specifically want to know about this case, of course.

Alternatively, we would have to implement it for the non-Windows certificate store case too, which would probably (have to) ignore some of the provided parameters. It is an operation we support, but we'll have to think about the caller side before implementing it.

kms/tpmkms/tpmkms.go
```diff
@@ -975,6 +981,54 @@ func (k *TPMKMS) storeCertificateChainToWindowsCertificateStore(req *apiv1.Store
...
+			deleteURI := uri.New("capi", url.Values{
+				"store-location": []string{storeLocation},
+				"store":          []string{store},
+				"issuer":         []string{issuer},
+				"serial":         []string{"0x" + cert.SerialNumber.Text(16)},
+			}).String()
+
+			if err := dk.DeleteCertificate(&apiv1.DeleteCertificateRequest{Name: deleteURI}); err != nil {
+				return fmt.Errorf("failed deleting expired certificate (serial %s): %w", cert.SerialNumber.Text(16), err)
+			}
```
hslatman, line 1031: Instead of returning immediately, this could continue looping, capturing all errors, and finally returning with `errors.Join`. That way the cleanup will be best-effort, and it could prevent erroneous cases from other other certs being removed.

kms/apiv1/options.go
```diff
@@ -76,6 +76,17 @@ type CertificateDeleter interface {
...
 
+// CleaningCertificateManager is an optional interface for KMS implementations
+// that support cleaning up expired certificates from a certificate store.
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
+// release.
+type CleaningCertificateManager interface {
+	Cleanup(issuer, storeLocation, store string, subjectRaw []byte) error
```
hslatman: The nature of this function maybe needs something like a struct to make it more flexible. That way it would become less specific to Windows. Most, if not all, other functions take a struct already. It could possibly be a URI-string API again, since some of those params are URI params already, but given the use case, I think a non-string API is better.

kms/tpmkms/tpmkms.go
```diff
@@ -1539,14 +1593,21 @@ type deletingCertificateManager interface {
 	DeleteCertificate(req *apiv1.DeleteCertificateRequest) error
 }
 
+type issuerCertificateFinder interface {
+	FindCertificatesByIssuer(req *apiv1.LoadCertificateRequest, subjectRaw []byte) ([]*x509.Certificate, error)
```
hslatman: Make similar changes where it is passed:
```suggestion
	FindCertificatesByIssuer(req *apiv1.LoadCertificateRequest, rawSubject []byte) ([]*x509.Certificate, error)
```

kms/tpmkms/tpmkms.go
darkfronza: Done: https://github.com/smallstep/crypto/pull/873/commits/10372d8a2eeebe1af1ca4ba39636ecd19da3a3a1

kms/tpmkms/tpmkms.go
darkfronza: Done: https://github.com/smallstep/crypto/pull/873/commits/3436ccb7ecc1183b418b956638e71e2027c848b0

kms/apiv1/options.go
darkfronza: Done: https://github.com/smallstep/crypto/pull/873/commits/6afec006073e6b15afa8c01cf7a9b0bb7f48530e

kms/tpmkms/tpmkms.go
darkfronza: Done: https://github.com/smallstep/crypto/pull/873/commits/9859f8bd8167d44045c0a9fcca8bc7b86d3f4be0

hslatman (dismissed):

hslatman (approved):

PR #885, opened 2025-10-30
Add support for providing `*tpm.TPM` instance when creating a `TPMKMS`

maraino:
This commits adds a new constructor to `tpmkms` that allows to initializes the `TPMKMS` with a given `*tpm.TPM`. instance.

Note that most of the file changes are because the new version of golangci-lint complains with legacy build tags, e.g. `// +build windows`

### DIFF
--- a/kms/tpmkms/tpmkms.go
+++ b/kms/tpmkms/tpmkms.go
@@ -199,9 +199,9 @@ func ParseTPMOptions(u *uri.URI) []tpm.NewTPMOption {
 	return opts
 }
 
-// ParseTPMKMSOptions is a helper method that returns a slice of [Option] for
+// ParseOptions is a helper method that returns a slice of [Option] for
 // the give URI.
-func ParseTPMKMSOptions(u *uri.URI) []Option {
+func ParseOptions(u *uri.URI) []Option {
 	opts := []Option{
 		WithAttestationCA(u.Get("attestation-ca-url"), u.Get("attestation-ca-root"), u.GetBool("attestation-ca-insecure")),
 		WithPermanentIdentifier(u.Get("permanent-identifier")), // TODO(hs): determine if this is needed
@@ -372,7 +372,7 @@ func New(ctx context.Context, opts apiv1.Options) (kms *TPMKMS, err error) {
 		}
 
 		tpmOpts = append(tpmOpts, ParseTPMOptions(u)...)
-		uriOptions = ParseTPMKMSOptions(u)
+		uriOptions = ParseOptions(u)
 	}
 
 	t, err := tpm.New(tpmOpts...)


kms/tpmkms/tpmkms.go
```diff
@@ -99,6 +96,137 @@ type TPMKMS struct {
...
+		o.windowsCNG = true
+		o.windowsIntermediateStore = store
+		o.windowsIntermediateStoreLocation = location
+		return nil
+	}
+}
+
+// TPMOptions is a helper method that returns a slice of [tpm.NewTPMOption] for
+// the given URI.
+func TPMOptions(u *uri.URI) []tpm.NewTPMOption {
```
hslatman: Slight change to the method name:
```suggestion
// ParseTPMOptions is a helper method that returns a slice of [tpm.NewTPMOption] for
// the given URI.
func ParseTPMOptions(u *uri.URI) []tpm.NewTPMOption {
```

An alternative could be `ExtractTPMOptions`

kms/tpmkms/tpmkms.go
```diff
@@ -99,6 +96,137 @@ type TPMKMS struct {
...
+	}
+	if storageDirectory := u.Get("storage-directory"); storageDirectory != "" {
+		opts = append(opts, tpm.WithStore(storage.NewDirstore(storageDirectory)))
+	}
+	return opts
+}
+
+// TPMKMSOptions is a helper method that returns a slice of [Option] for the
+// give URI.
+func TPMKMSOptions(u *uri.URI) []Option {
```
hslatman: Since this is in the `tpmkms` package, I think we can shorten it to this:
```suggestion
// ParseOptions is a helper method that returns a slice of [Option] for the
// give URI.
func ParseOptions(u *uri.URI) []Option {
```

kms/tpmkms/tpmkms_test.go
```diff
@@ -40,13 +75,108 @@ func TestNew(t *testing.T) {
...
 			}
 		})
 	}
 }
 
+func TestNew_no_windows(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("This test cannot run on windows")
+	}
+
```
hslatman: It looks like the intention is to only run the test on other OSs, because it's expected to return an error and `nil`? Instead of skipping it, would it work to put this in `tpmkms_others_test.go`?

kms/tpmkms/tpmkms_test.go
```diff
@@ -40,13 +75,108 @@ func TestNew(t *testing.T) {
...
+			tt.assertion(t, err)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func TestNewWithTPM_no_windows(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("This test cannot run on windows")
+	}
```
hslatman: Idem

kms/tpmkms/tpmkms.go
```diff
@@ -221,93 +349,84 @@ const (
...
+	t, err := tpm.New(tpmOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed creating new TPM: %w", err)
+	}
+
+	return NewWithTPM(ctx, t, uriOptions...)
+}
+
+// NewWithTPM initializes a new KMS backed by the given TPM.
+func NewWithTPM(ctx context.Context, t *tpm.TPM, opts ...Option) (*TPMKMS, error) {
```
hslatman, line 392: Maybe mark this as experimental, as I can see us _potentially_ accepting a TPM interface in the future too, and assuming we don't want another new function for that.

kms/tpmkms/tpmkms.go
```diff
@@ -99,6 +96,137 @@ type TPMKMS struct {
 	identityEarlyRenewalEnabled      bool
 }
 
+// Option is the type used as a variadic argument in NewWithTPM.
+type Option func(o *options) error
```
hslatman, line 105: _If_ the `NewWithTPM` function is marked experimental, this probably also should be marked like that.

kms/tpmkms/tpmkms_windows_test.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := New(tt.args.ctx, tt.args.opts)
+			tt.assertion(t, err)
+			if got != nil {
+				assert.NotNil(t, got.tpm)
+				assert.IsType(t, &tpm.TPM{}, got.tpm)
+				tt.want.tpm = got.tpm
+			}
```
hslatman, line 101: Maybe this needs a short comment to explain why the expected value is purposefully altered with the value received.

Alternatively, the equality assertion below could be altered, but that'll need some additional assertions then.

hslatman (commented):
Looking pretty nice overall! Looking forward to try it 😄

kms/tpmkms/tpmkms.go
maraino: Fixed with 2ef8141

kms/tpmkms/tpmkms.go
maraino: Fixed with 8f1cab7

kms/tpmkms/tpmkms_test.go
maraino: Fixed with 2ef8141

kms/tpmkms/tpmkms_test.go
maraino: Fixed with 2ef8141

kms/tpmkms/tpmkms.go
maraino: Fixed with 2ef8141

kms/tpmkms/tpmkms.go
maraino: Fixed with 2ef8141

PR #910, opened 2025-12-13
Platform KMS

maraino:
This is a first attempt to create a platform KMS. This KMS is supports mackms, tpmkms, softkms, and capikms.

There's a couple of things that I'll like to consider:
1. Is `hw` the right argument, or should we use `ak`?
2. Should we fail if `hw` (or `ak`) is set for softkms and capikms?

### DIFF
--- a/kms/apiv1/requests.go
+++ b/kms/apiv1/requests.go
@@ -268,11 +268,21 @@ type AttestationClient interface {
 type attestSignerCtx struct{}
 
 // NewAttestSignerContext creates a new context with the given signer.
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
+// release.
 func NewAttestSignerContext(ctx context.Context, signer crypto.Signer) context.Context {
 	return context.WithValue(ctx, attestSignerCtx{}, signer)
 }
 
 // AttestSignerFromContext returns the signer from the context.
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
+// release.
 func AttestSignerFromContext(ctx context.Context) (crypto.Signer, bool) {
 	signer, ok := ctx.Value(attestSignerCtx{}).(crypto.Signer)
 	return signer, ok
--- a/kms/platform/kms_darwin.go
+++ b/kms/platform/kms_darwin.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"maps"
 	"net/url"
+	"strings"
 
 	"go.step.sm/crypto/kms/apiv1"
 	"go.step.sm/crypto/kms/mackms"
@@ -72,7 +73,7 @@ func transformToMacKMS(rawuri string) (string, error) {
 		if !u.uri.Has("keychain") {
 			uv.Set("keychain", "dataProtection")
 		}
-	} else if u.uri.Has("hw") {
+	} else if strings.EqualFold(u.uri.Get("hw"), "false") {
 		uv.Set("se", "false")
 	}
 
--- a/kms/platform/kms_tpmsimulator_test.go
+++ b/kms/platform/kms_tpmsimulator_test.go
@@ -3,7 +3,6 @@
 package platform
 
 import (
-	"context"
 	"crypto"
 	"crypto/x509"
 	"crypto/x509/pkix"
@@ -57,7 +56,7 @@ func mustTPMDevice(t *testing.T) (*tpm.TPM, string, string) {
 
 	listener := &net.ListenConfig{}
 	socket := filepath.Join(dir, "tpm.sock")
-	ln, err := listener.Listen(context.TODO(), "unix", socket)
+	ln, err := listener.Listen(t.Context(), "unix", socket)
 	require.NoError(t, err)
 
 	go func() {
--- a/kms/tpmkms/tpmkms.go
+++ b/kms/tpmkms/tpmkms.go
@@ -1078,7 +1078,6 @@ func (k *TPMKMS) deleteCertificateFromWindowsCertificateStore(req *apiv1.DeleteC
 	if err := dk.DeleteCertificate(&apiv1.DeleteCertificateRequest{
 		Name: uri.New("capi", uv).String(),
 	}); err != nil {
-		fmt.Println(uri.New("capi", uv).String())
 		return fmt.Errorf("failed deleting certificate using Windows platform cryptography provider: %w", err)
 	}
 
--- a/kms/platform/kms_darwin_test.go
+++ b/kms/platform/kms_darwin_test.go
@@ -28,8 +28,10 @@ func Test_transformToMacKMS(t *testing.T) {
 		{"scheme", "kms:", "mackms:", assert.NoError},
 		{"with name", "kms:name=foo", "mackms:label=foo", assert.NoError},
 		{"with hw", "kms:name=foo;hw=true", "mackms:keychain=dataProtection;label=foo;se=true", assert.NoError},
+		{"with hw false", "kms:name=foo;hw=false", "mackms:label=foo;se=false", assert.NoError},
 		{"with hw on query", "kms:name=foo?hw=true", "mackms:keychain=dataProtection;label=foo;se=true", assert.NoError},
 		{"with hw and keychain", "kms:name=foo;hw=true;keychain=my", "mackms:keychain=my;label=foo;se=true", assert.NoError},
+		{"with hw other", "kms:name=foo;hw=other", "mackms:label=foo", assert.NoError},
 		{"with extrasValues", "kms:name=foo;keychain=my?foo=bar&baz=qux", "mackms:baz=qux;foo=bar;keychain=my;label=foo", assert.NoError},
 		{"fail parse", "softkms:name=foo", "", assert.Error},
 	}


kms/apiv1/options.go
```diff
@@ -151,6 +173,9 @@ const (
 	TPMKMS Type = "tpmkms"
 	// MacKMS is the KMS implementation using macOS Keychain and Secure Enclave.
 	MacKMS Type = "mackms"
+	// PlatformKMS is the KMS implementation that uses TPMKMS on Windows and
+	// Linux and MacKMS on macOS..
```
hslatman:
```suggestion
	// Linux and MacKMS on macOS.
```

kms/apiv1/options.go
```diff
@@ -18,6 +18,17 @@ type KeyManager interface {
...
+// deleting keys.
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later
+// release.
+type KeyDeleter interface {
+	DeleteKey(req *DeleteKeyRequest) error
+}
+
```
hslatman, line 31: When considering/introducing these new interfaces, we may want to add a ctx as the first argument to make them more future-proof.

The main reason is I foresee us refactoring existing KMS implementations and the requests to take one too, so that we can cancel requests to remote KMSs and carry OpenTelemetry data, for example.

If we really want to keep the old function signatures we could pass a `ctx` in the requests, but that's not idiomatic Go. It might be an option to do that temporarily, so that not everything breaks, but then we'd have to work on the same parts at least twice.

kms/capi/ncrypt_windows.go
```diff
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"unsafe"
 
+	"go.step.sm/crypto/kms/apiv1"
 	"golang.org/x/sys/windows"
 )
```
hslatman:
```suggestion
    "golang.org/x/sys/windows"

    "go.step.sm/crypto/kms/apiv1"
)

```

kms/mackms/mackms.go
```diff
@@ -1276,17 +1276,15 @@ func parseCertURI(rawuri string, useDataProtectionKeychain, requireValue bool) (
...
+	serialNumber, err := u.GetBigInt("serial")
+	if err != nil {
+		return nil, fmt.Errorf("error parsing %q: %w", rawuri, err)
 	}
-
-	var serialNumber *big.Int
-	if len(serial) > 0 {
-		serialNumber = new(big.Int).SetBytes(serial)
+	if requireValue && label == "" && serialNumber == nil {
+		return nil, fmt.Errorf("error parsing %q: label or serial are required", rawuri)
```
hslatman: Either this:
```suggestion
	if requireValue && label == "" && serialNumber == nil {
		return nil, fmt.Errorf("error parsing %q: label or serial is required", rawuri)
```
or something like "one of label or serial is required".

kms/platform/kms.go
```diff
@@ -0,0 +1,258 @@
...
+	extraValues := make(url.Values)
+	for k, v := range u.Values {
+		if k != nameKey && k != hwKey && k != backendKey {
+			extraValues[k] = v
+		}
+	}
+
+	return &kmsURI{
+		uri:         u,
+		backend:     apiv1.Type(strings.ToLower(u.Get(backendKey))),
```
hslatman, line 58: Probably good to validate that `backendKey` points to one of the supported types before setting it.

kms/platform/kms.go
```diff
@@ -0,0 +1,258 @@
...
+	})
+}
+
+func (k *KMS) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyResponse, error) {
+	name, err := k.transform(req.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	req = clone(req)
```
hslatman: It might be clearer, and possibly safer, to assign to a new variable (for all cases). It clones the value, but overrides the pointer, so if for some reason the caller does something with the request _after_ calling the function, it would look at something different.

kms/platform/kms_darwin.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	u, err := parseURI(opts.URI)
+	if err != nil {
+		return nil, err
+	}
+
+	switch u.backend {
+	case apiv1.TPMKMS:
+		return newTPMKMS(ctx, opts)
```
hslatman, line 29: This is just for testing purposes? I'm OK with including it, but it might be nicer if there's an easy way to only enable it for `go run|test` and dev builds. Not a blocker, though.

kms/platform/kms_test.go
```diff
@@ -0,0 +1,1121 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.step.sm/crypto/keyutil"
+	"go.step.sm/crypto/kms/apiv1"
+	"go.step.sm/crypto/kms/uri"
+	"go.step.sm/crypto/minica"
+	"go.step.sm/crypto/pemutil"
+	"go.step.sm/crypto/randutil"
+	"go.step.sm/crypto/tpm"
+)
```
hslatman, line 33:
```suggestion
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
	
	"go.step.sm/crypto/keyutil"
	"go.step.sm/crypto/kms/apiv1"
	"go.step.sm/crypto/kms/uri"
	"go.step.sm/crypto/minica"
	"go.step.sm/crypto/pemutil"
	"go.step.sm/crypto/randutil"
	"go.step.sm/crypto/tpm"
)
```

kms/platform/kms_test.go
```diff
@@ -0,0 +1,1121 @@
...
+	os.Exit(m.Run())
+}
+
+func isTPMAvailable() bool {
+	t, err := tpm.New()
+	if err != nil {
+		return false
+	}
+	return t.Available() == nil
+}
```
hslatman: This could use the `go.step.sm/tpm/available.Check` function.

kms/platform/kms_tpm.go
```diff
@@ -0,0 +1,99 @@
...
+func transformToTPMKMS(u *kmsURI) string {
+	uv := url.Values{}
+	if u.name != "" {
+		uv.Set("name", u.name)
+	}
+	if u.hw {
+		uv.Set("ak", "true")
+	} else if u.uri.Has("hw") {
+		uv.Set("ak", "false")
+	}
```
hslatman: I think this logic needs to be removed. 

The decision to set `ak=true|false` does not depend on `hw` being set. `ak=true` is for creating/listing/getting/deleting attestation keys. When `ak=false`, the same operations are for generic private keys backed by the TPM.

kms/platform/kms_tpm.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Add custom extra values that might be tpmkms specific.
+	maps.Copy(uv, u.extraValues)
```
hslatman, line 72: I'm wondering if it would be good to validate that there's only supported key/value pairs for the specific KMS that we're in, and fail if something's off. Or will it result in a fail one level deeper?

If updating here, also check the other KMSs.

kms/platform/kms_tpm.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	uv := url.Values{
+		"name": []string{u.Get("name")},
+	}
+	if u.GetBool("ak") {
+		uv.Set("hw", "true")
+	}
```
hslatman: See the other comment regarding `ak=true|false`.

kms/platform/kms_windows.go
```diff
@@ -0,0 +1,86 @@
...
+
+	u, err := parseURI(opts.URI)
+	if err != nil {
+		return nil, err
+	}
+
+	switch u.backend {
+	case apiv1.CAPIKMS:
+		opts.URI = transformToCapiKMS(u)
+		return newCAPIKMS(ctx, opts)
```
hslatman: Instead of assigning to `opts.URI`, call `transformToCapiKMS` in `newCAPIKMS`?

Also, capitalization of CAPI can be made consistent.

kms/platform/kms_windows.go
```diff
@@ -0,0 +1,86 @@
...
+
+func transformToCapiKMS(u *kmsURI) string {
+	uv := url.Values{}
+	if u.name != "" {
+		uv.Set("key", u.name)
+	}
+
+	// When storing certificate skip key validation.
+	// This avoid a prompt looking for an SmartCard.
+	uv.Set("skip-find-certificate-key", "true")
```
hslatman: Should this happen for all calls? It is now done unconditionally.

kms/softkms/softkms_test.go
```diff
@@ -17,7 +17,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-
+	"github.com/stretchr/testify/require"
 	"go.step.sm/crypto/kms/apiv1"
```
hslatman, line 22:
```suggestion
	"github.com/stretchr/testify/require"
	
        "go.step.sm/crypto/kms/apiv1"
```

kms/uri/uri.go
```diff
@@ -218,6 +251,45 @@ func (u *URI) Read(key string) ([]byte, error) {
 	return readFile(path)
 }
 
+// hexString returns a clean hexadecimal string and a boolean indicating if s
+// can be an hexadecimal string. If s starts with 0x (0x12), 0X (0X1A), or
+// contains colons (01:1A) it will remove them and return true if it only
+// contains valid hexadecimal characters. It will also true if the string
```
hslatman:
```suggestion
// contains valid hexadecimal characters. It will also return true if the string
```

kms/uri/uri.go
```diff
@@ -218,6 +251,45 @@ func (u *URI) Read(key string) ([]byte, error) {
...
+// contains at least one letter A-F (010A). It will also prefix the string with
```
hslatman:
```suggestion
// contains at least one letter A-F (010A). It will prefix the string with
```

kms/platform/kms_windows_test.go
```diff
@@ -0,0 +1,499 @@
...
+	"crypto/x509"
+	"encoding/hex"
+	"fmt"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.step.sm/crypto/kms/apiv1"
+	"go.step.sm/crypto/kms/uri"
```
hslatman, line 20:
```suggestion
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
	
	"go.step.sm/crypto/kms/apiv1"
	"go.step.sm/crypto/kms/uri"
```

kms/capi/ncrypt_windows.go
maraino: Fixed with 3b94424

kms/platform/kms_test.go
maraino: Fixed with 3b94424

kms/softkms/softkms_test.go
maraino: Fixed with 3b94424

kms/platform/kms_windows_test.go
maraino: Fixed with 3b94424

kms/platform/kms.go
maraino: I will change it, but in most of the cases is safe, the caller will have a reference to the old pointer and we will have a reference to a new pointer. The caller will see the original name.

But as we are not deep cloning all types, if the request has a pointer, this one can change and it could be a problem.

kms/platform/kms_darwin.go
maraino: It would be hard to enforce this here, but the program using this package can.

kms/platform/kms_windows.go
maraino: I will capitalize the methods, but I did the transformation there to avoid parsing the URI again.

kms/platform/kms_windows.go
maraino: Only for those that store certificates, but right now I don't have different methods or an argument indicating which kind of transformation we want. This could be an improvement but I think this is the only use case for it.

kms/apiv1/options.go
maraino: Fixed with 49f2331

kms/mackms/mackms.go
maraino: Fixed with 49f2331

kms/platform/kms.go
maraino: Fixed with 49f2331

kms/platform/kms_test.go
maraino: Fixed with 49f2331

kms/platform/kms_tpm.go
maraino: Controlling all the parameters here is way more complex. If we add new ones, we would have to control those here, so parameters like `storage-directory` on TPMs or `provider` on CAPI would be managed on the backend KMS.

kms/platform/kms_windows.go
maraino: Fixed with 49f2331

kms/uri/uri.go
maraino: Fixed with 49f2331

kms/uri/uri.go
maraino: Fixed with 49f2331

kms/apiv1/options.go
maraino: We are already adding an interface like this in other places, here I'm making it public. If we want to add the context, I think we could add It everywhere, using one of this options:
1. ctx argument in functions, perhaps using `apiv2`
2. Context field in requests
3. Something like `http.NewRequestWithContext` that sets an private field, that can be accessed by a `Context()` method, that's more or less like `2` the only differences is that in that method we can return `context.Background()` if it's not set instead of adding if conditions in all the methods where contexts are used.

kms/platform/kms.go
maraino: `backend` is only used when a new KMS is initialized and it is validated there, depending on the OS. I don't think is necessary duplicate the function to make it more explicit.

maraino:
@hslatman I've fixed most of your comments except the ones related to `hw` and `ak`. I see a few options:

1. Change `hw` with `ak` and use `se=true` if `ak=true` (we're doing this in other places)
2. Keep `hw` and add `ak`. `hw` will be ignored in TPMKMS, but will be used in `mackms` for `se=true` and `capikms` for `provider=Microsoft+Platform+Crypto+Provider`
3. My least favorite, remove `hw` and add `ak` on TPMKMS. And use backend parameters `se` and `provider`.

maraino:
@hslatman What do you think about the custom attestation using the AttestationClient in f1076d3

kms/platform/kms.go
hslatman: Ah yea; my bad. I didn't realize it was mostly safe. Anyhow, thanks 😄

kms/apiv1/options.go
hslatman: I would go with option 1. I'd say it would be OK to keep `apiv1`, as the request types themselves don't change or get a different meaning, or something like that.

hslatman: Let's keep it like this for now, considering that the interface already exists. We can add `ctx` later.

kms/apiv1/requests.go
```diff
@@ -265,6 +265,19 @@ type AttestationClient interface {
...
 
+type attestSignerCtx struct{}
+
+// NewAttestSignerContext creates a new context with the given signer.
+func NewAttestSignerContext(ctx context.Context, signer crypto.Signer) context.Context {
+	return context.WithValue(ctx, attestSignerCtx{}, signer)
+}
+
+// AttestSignerFromContext returns the signer from the context.
+func AttestSignerFromContext(ctx context.Context) (crypto.Signer, bool) {
```
hslatman, line 286: To be safe, we could mark these as experimental, but not a blocker.

kms/tpmkms/tpmkms.go
```diff
@@ -1071,12 +1078,23 @@ func (k *TPMKMS) deleteCertificateFromWindowsCertificateStore(req *apiv1.DeleteC
 	if err := dk.DeleteCertificate(&apiv1.DeleteCertificateRequest{
 		Name: uri.New("capi", uv).String(),
 	}); err != nil {
+		fmt.Println(uri.New("capi", uv).String())
```
hslatman:
```suggestion
```

kms/platform/kms_darwin.go
```diff
@@ -0,0 +1,106 @@
...
+		uv.Set("label", u.name)
+	}
+	if u.hw {
+		uv.Set("se", "true")
+		if !u.uri.Has("keychain") {
+			uv.Set("keychain", "dataProtection")
+		}
+	} else if u.uri.Has("hw") {
+		uv.Set("se", "false")
+	}
```
hslatman: Should this explicitly check for the value to be false? I think that makes it clearer.

kms/platform/kms_tpmsimulator_test.go
```diff
@@ -0,0 +1,906 @@
...
+	})
+	require.NoError(t, sim.Open())
+
+	dir := t.TempDir()
+	stpm, err := tpm.New(tpm.WithSimulator(sim), tpm.WithStore(storage.NewDirstore(dir)))
+	require.NoError(t, err)
+
+	listener := &net.ListenConfig{}
+	socket := filepath.Join(dir, "tpm.sock")
+	ln, err := listener.Listen(context.TODO(), "unix", socket)
```
hslatman: I think this'll work:
```suggestion
	ln, err := listener.Listen(t.Context(), "unix", socket)
```

kms/apiv1/requests.go
maraino: Fixed with cef679f

kms/tpmkms/tpmkms.go
maraino: Fixed with cef679f

kms/platform/kms_darwin.go
maraino: Fixed with cef679f

kms/platform/kms_tpmsimulator_test.go
maraino: Fixed with cef679f

PR #916, opened 2025-12-15
Add certificate chain support for capi

maraino:
This commit moves the certificate chain support from the tpmkms to the capi kms.

### DIFF
--- a/kms/capi/capi.go
+++ b/kms/capi/capi.go
@@ -11,8 +11,6 @@ import (
 	"crypto/elliptic"
 	"crypto/rsa"
 	"crypto/x509"
-	"crypto/x509/pkix"
-	stdans1 "encoding/asn1"
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
@@ -25,13 +23,15 @@ import (
 	"unsafe"
 
 	"github.com/pkg/errors"
+
+	"golang.org/x/crypto/cryptobyte"
+	"golang.org/x/crypto/cryptobyte/asn1"
+	"golang.org/x/sys/windows"
+
 	"go.step.sm/crypto/fingerprint"
 	"go.step.sm/crypto/kms/apiv1"
 	"go.step.sm/crypto/kms/uri"
 	"go.step.sm/crypto/randutil"
-	"golang.org/x/crypto/cryptobyte"
-	"golang.org/x/crypto/cryptobyte/asn1"
-	"golang.org/x/sys/windows"
 )
 
 // Scheme is the scheme used in uris, the string "capi".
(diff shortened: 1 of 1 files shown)


kms/capi/capi.go
```diff
@@ -32,24 +38,32 @@ import (
...
+	KeySpec                      = "key-spec"                  // 0, 1, 2; none/NONE, at_keyexchange/AT_KEYEXCHANGE, at_signature/AT_SIGNATURE
+	SkipFindCertificateKey       = "skip-find-certificate-key" // skips looking up certificate private key when storing a certificate
 )
 
 const (
 	MachineStore = "machine"
 	UserStore    = "user"
+	MyStore      = "My"
+	CAStore      = "CA" // TODO(hs): verify "CA" works for "machine" certs too
 )
```
hslatman, line 66: There's a distinction between stores and store locations. For clarity `Location` can be suffixed to `machine` and `user`.

maraino: Yes, I didn't want to change the exported values, but I can add a deprecated value.

maraino: Fixed with 10afab6

kms/capi/capi.go
```diff
@@ -4,22 +4,28 @@ package capi
...
 	"bytes"
+	"cmp"
 	"context"
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rsa"
 	"crypto/x509"
+	"crypto/x509/pkix"
+	stdans1 "encoding/asn1"
```
hslatman:
```suggestion
	stdasn1 "encoding/asn1"
```

kms/capi/capi.go
```diff
@@ -1041,4 +1151,24 @@ func (s *CAPISigner) Public() crypto.PublicKey {
 	return s.PublicKey
 }
 
+type subjectPublicKeyInfo struct {
+	Algorithm        pkix.AlgorithmIdentifier
+	SubjectPublicKey stdans1.BitString
```
hslatman:
```suggestion
	SubjectPublicKey stdans1.BitString
```

kms/capi/capi.go
```diff
@@ -60,6 +79,59 @@ var signatureAlgorithmMapping = map[apiv1.SignatureAlgorithm]string{
 	apiv1.ECDSAWithSHA512:          ALG_ECDSA_P521,
 }
 
+type uriAttributes struct {
+	ContainerName             string
```
hslatman: The properties in this struct can likely become private too

kms/capi/capi.go
```diff
@@ -703,30 +742,71 @@ func (k *CAPIKMS) LoadCertificate(req *apiv1.LoadCertificateRequest) (*x509.Cert
...
+				break
+			}
+			return nil, fmt.Errorf("failed loading intermediate CA certificate using Windows platform cryptography provider: %w", err)
+		}
+
+		// if the discovered parent has a signature from itself, assume it's a root CA,
+		// and break from the loop
+		if parent.CheckSignatureFrom(parent) == nil {
+			break
+		}
```
hslatman, line 782: Not for this PR, but just FYI: there's a new draft RFC proposing "Unsigned Certificates": https://datatracker.ietf.org/doc/draft-ietf-lamps-x509-alg-none/. Its primary use case is for root CA certificates, for which the self-signature doesn't provide functional value. It will allow for (overall) smaller certificate chains by not having the root CA signature, which should be helpful especially for PQC chains.

In places where we currently depend on this self-signature check to determine if a cert is a root, we should update it to take that new algorithm into account. It probably also needs some changes on the Go side.

kms/capi/capi.go
maraino: Fixed in [78976dc](https://github.com/smallstep/crypto/pull/916/commits/78976dc6f3e38ea7ba76fd932d8077172edbc843), it turns out that this is no longer needed.

kms/capi/capi.go
maraino: Fixed in [78976dc](https://github.com/smallstep/crypto/pull/916/commits/78976dc6f3e38ea7ba76fd932d8077172edbc843), it turns out that this is no longer needed.

kms/capi/capi.go
maraino: Fixed in [78976dc](https://github.com/smallstep/crypto/pull/916/commits/78976dc6f3e38ea7ba76fd932d8077172edbc843).

PR #1006, opened 2026-04-24
Add store-location=machine URI option for Windows machine key store

darkfronza:
Introduces WithMachineKey() TPM option wired to the store-location=machine URI parameter in tpmkms. On Windows, this causes NCRYPT_MACHINE_KEY_FLAG to be passed to NCrypt key operations, directing key creation and access to the local machine key store rather than the current user key store. Has no effect on non-Windows platforms.

Also bumps go-attestation from pre-release v0.4.4-... to released v0.4.5, which introduces the MachineKey field used by this option.


tpm/tpm.go
```diff
@@ -96,6 +96,17 @@ func WithCommandChannel(commandChannel CommandChannel) NewTPMOption {
...
+// (local machine) key store rather than the current user key store. On Windows
+// this causes NCRYPT_MACHINE_KEY_FLAG to be passed to NCrypt key operations.
+// This option has no effect on non-Windows platforms.
+func WithMachineKey() NewTPMOption {
+	return func(o *options) error {
+		o.attestConfig.MachineKey = true
+		return nil
+	}
+}
+
```
hslatman, line 109: This option will affect operations on all keys going through the specific `tpm.TPM` instance it is applied on. It might be OK, if separate `tpm.TPM` instances are used, but I was wondering if the alternative was considered that passes this config for specific keys.

tpm/tpm.go
```diff
@@ -241,6 +252,7 @@ func (t *TPM) initializeCommandChannel() error {
 		t.attestConfig = &attest.OpenConfig{
 			TPMVersion:     t.options.attestConfig.TPMVersion,
 			CommandChannel: t.commandChannel,
+			MachineKey:     t.options.attestConfig.MachineKey,
```
hslatman, line 255: The option probably shouldn't sit on `attestConfig`. That's used for higher level config of opening a TPM in a specific mode / configuration. Can probably sit on `options` directly, since it's just passed around.

tpm/ak.go
```diff
@@ -298,7 +300,7 @@ func (t *TPM) DeleteAK(ctx context.Context, name string) (err error) {
 		return fmt.Errorf("failed persisting storage: %w", err)
 	}
 
-	return
+	return attestErr
```
hslatman, line 303: Is the goal to be able to act on this error case? A typed error might help.

tpm/key.go
```diff
@@ -398,7 +403,7 @@ func (t *TPM) DeleteKey(ctx context.Context, name string) (err error) {
 		return fmt.Errorf("failed persisting storage: %w", err)
 	}
 
-	return
+	return attestErr
```
hslatman, line 406: See the remark for the other case

tpm/internal/key/pcp_windows.go
```diff
@@ -491,10 +500,13 @@ func decodeKeyBlob(keyBlob []byte) ([]byte, []byte, error) {
 }
 
 // openPCP initializes a reference to the Microsoft PCP provider.
+// Pass machineKey=true to create and open keys in the machine (local machine)
+// key store rather than the current user key store.
 // The Caller is expected to call Close() when they are done.
-func openPCP() (*winPCP, error) {
+func openPCP(machineKey bool) (*winPCP, error) {
```
hslatman, line 506: Instead of passing this via `openPCP`, it looks better to me if passed via the `KeyConfig`.

PR #1021, opened 2026-05-09
Add per-key key-scope option for Windows machine vs user key store

joshdrake:
## Summary

- Adds `key-scope=machine|user` URI parameter to `kms/tpmkms`, controlling whether the underlying private key lives in the local machine key store (`NCRYPT_MACHINE_KEY_FLAG`) or the current user's key store. Independent of `store-location`, so callers can pair a machine-store certificate with a user-owned key (e.g., a `LocalSystem` service that wants its cert visible machine-wide but its key only openable by `SYSTEM`) or vice versa.
- Threads `MachineKey` per-key through `tpm.CreateKeyConfig` / `tpm.AttestKeyConfig` / `tpm.CreateAKConfig` and into the underlying NCrypt PCP create call. **Per-key, not per-TPM-instance** — a single `tpm.TPM` handle can mint or open keys in either scope.
- `t.open()` reads the requested scope from context and applies it to `attestConfig.MachineKey` under the lock just before `attest.OpenTPM`. Safe because `attestTPM` is recreated on every `open()`.
- Persists `MachineKey` in `storage.AK` and `storage.Key` so reload after process restart uses the matching scope without the URI needing to carry it. JSON `omitempty` so existing on-disk stores remain readable and default to `false` (= existing user-scope behavior).
- `tpmkms.CreateKey` preserves `key-scope=machine` in the returned URI so downstream `CreateSigner`/`DeleteKey` calls don't lose scope (closes a returned-URI bug we hit during testing).
- `AttestKey` reads the AK's persisted scope and uses it for the attest session — attested keys inherit their AK's scope; conflicting `AttestKeyConfig.MachineKey` is rejected with a clear error.
- `DeleteKey`/`DeleteAK` always clean up the file-store entry even if the in-TPM (NCrypt) deletion fails, preventing repeated retries from re-encountering the same failing entry. When the file-store cleanup succeeded but the in-TPM step did not, returns a typed `*PartialDeleteError` so callers can distinguish a partial delete from total failure.
- Bumps `github.com/smallstep/go-attestation` to `v0.4.9` for the `OpenConfig.MachineKey` field this change drives.

Defaults: when `key-scope` is unset, derive from `store-location` for back-compat (`store-location=machine` → `key-scope=machine`).

## Background

Code review and a Windows test matrix (interactive admin + `LocalSystem` via `psexec -s -i 0`, both x86_64) showed that the agent's TPM-backed device-identity keys land under `LocalSystem`'s user-profile PCPKSP scope rather than the machine scope, even when `store-location=machine` is set. The `kms/capi` machine-key fix from #802 doesn't apply because the agent uses `kms/tpmkms` (with `enable-cng=true`) — that path goes through `tpm/internal/key/pcp_windows.go`, where `nCryptCreatePersistedKey` was hardcoded to `dwFlags=0`. Fixing it requires (a) plumbing a `MachineKey` flag through the `tpm` package and (b) deciding how to express user intent at the URI layer.

This PR is an alternative shape to #1006. Differences:

- **Per-key, not per-TPM-instance.** Herman's review feedback on #1006 pushed back on `WithMachineKey()` being a per-TPM option (would affect all keys going through that handle); the discussion converged on per-key being the right model. This PR puts `MachineKey` on the per-key configs (`CreateKeyConfig`, `AttestKeyConfig`, `CreateAKConfig`) and uses a context value to thread it into `t.open()` for the operations that need it.
- **`MachineKey` lives on `options`, not `attestConfig`.** Also from Herman's review.
- **`key-scope` is decoupled from `store-location`.** Cert location and key ownership are orthogonal Windows concepts. There are real scenarios where a service wants its cert in `LocalMachine\My` for visibility but its key restricted to one identity (e.g., `LocalSystem`-only) — `key-scope=user` with `store-location=machine` expresses that. Default-derive from `store-location` for back-compat.
- **Typed `*PartialDeleteError` for `DeleteKey`/`DeleteAK`.** Also from Herman's review.

## Test plan

- [x] `go build ./...` clean on `linux/amd64` and `windows/amd64`
- [x] `go vet ./...` clean (one pre-existing master warning in `kms/capi/ncrypt_windows.go:569`, unrelated to this change)
- [x] `go test ./...` green across all packages
- [x] **Empirical Windows matrix** (throwaway diagnostic harness, run interactive admin + `psexec -s -i 0` LocalSystem):
  - `key-scope=machine` lands keys at `C:\ProgramData\Microsoft\Crypto\PCPKSP\<sid-hash>\<file>.PCPKEY` (interactive admin + LocalSystem).
  - `key-scope=user` lands keys at `<userprofile>\AppData\Local\Microsoft\Crypto\PCPKSP\<sid-hash>\<file>.PCPKEY` (LocalSystem profile when running as SYSTEM).
  - **Decoupling proven:** `key-scope=machine` + `store-location=user` puts cert in user store and key in machine PCPKSP; `key-scope=user` + `store-location=machine` puts cert in machine store and key in user PCPKSP.
  - Back-compat: `key-scope` unset + `store-location=machine` correctly resolves to machine scope.
  - `CreateSigner` via the URI returned by `CreateKey` round-trips successfully in every TPM scenario; sign+verify ok.

## Out of scope

- Agent-side change to set `key-scope=machine` on the device-identity TPM URI; lands separately in the agent repo.
- Equivalent fix for `kms/capi`'s returned-URI bug (different root cause, different file). Tracked separately.


kms/tpmkms/uri.go
```diff
@@ -27,6 +27,28 @@ type objectProperties struct {
...
+	scope := o.keyScope
+	if scope == "" {
+		if o.storeLocation == "machine" {
+			scope = "machine"
+		} else {
+			scope = "user"
+		}
+	}
+	return scope == "machine"
 }
```
smst-jeff:
```suggestion
// isMachineKey returns true if the resolved key scope is "machine".
// When keyScope is unset on the object, the value is derived from
// storeLocation: "machine" → true, anything else → false.
func (o objectProperties) isMachineKey() bool {
	return o.keyScope == "machine" ||
		(o.keyScope == "" && o.storeLocation == "machine")
}
```

tpm/key.go
```diff
@@ -236,7 +251,32 @@ func (w attestValidationWrapper) Validate() error {
...
+	// To opt into "use whatever scope the AK is in," leave
+	// AttestKeyConfig.MachineKey at its zero value AND make sure the AK
+	// is in user scope; for a machine-scope AK, set MachineKey: true
+	// explicitly.
+	if config.MachineKey != ak.MachineKey {
+		return nil, fmt.Errorf("AttestKeyConfig.MachineKey=%v does not match AK %q scope (MachineKey=%v); attested keys inherit their AK's scope", config.MachineKey, akName, ak.MachineKey)
+	}
+	machineKey := ak.MachineKey
+
+	if err = t.open(withMachineKey(ctx, machineKey)); err != nil {
```
hslatman: Changing the order of `t.open` (which calls into `t.store.load`) and `t.store.X` could result in unexpected behavior depending on the `Storage` implementation. With the `Dirstore` implementation it won't be an issue in practice, but for other implementations it can be different.

Maybe `t.store.Load` can be extracted from `t.open`, and be called before any `t.store` operation.

kms/tpmkms/uri.go
joshdrake: 5f2cf52

tpm/key.go
joshdrake: [c3de912](https://github.com/smallstep/crypto/pull/1021/commits/c3de912e714a494b98078e44ce7cd982e68b7e7a)

I simply added an explicit `t.store.Load` for now to preserve the order. There will be duplicate `Load` calls, but I figured that's probably okay for now to hold off on refactoring `open`.

tpm/ak.go
```diff
@@ -133,8 +142,16 @@ func (ak *AK) MarshalJSON() ([]byte, error) {
...
+// To request a machine-scoped AK on Windows, use [TPM.CreateAKWithConfig].
 func (t *TPM) CreateAK(ctx context.Context, name string) (ak *AK, err error) {
-	if err = t.open(ctx); err != nil {
+	return t.CreateAKWithConfig(ctx, name, CreateAKConfig{})
+}
+
+// CreateAKWithConfig creates and stores a new AK with the given config.
+// See [TPM.CreateAK] for the simpler signature.
+func (t *TPM) CreateAKWithConfig(ctx context.Context, name string, config CreateAKConfig) (ak *AK, err error) {
+	if err = t.open(withMachineKey(ctx, config.MachineKey)); err != nil {
```
maraino: The use of the context to store a specific configuration doesn't look natural. As this is a private method we can add a structure, similar to the `CreateAKConfig` here when opening the TPM. We can also use variadic methods changing an options structure.

tpm/tpm.go
```diff
@@ -190,6 +190,12 @@ func (t *TPM) open(ctx context.Context) (err error) {
 		return fmt.Errorf("failed initializing command channel: %w", err)
 	}
 
+	// Apply per-call machine-key scope to the attest open config. We hold
+	// t.lock at this point and t.attestTPM is re-created on every open, so
+	// mutating attestConfig here is safe and does not leak across callers.
+	// Always set explicitly so we never carry a previous caller's choice.
+	t.attestConfig.MachineKey = machineKeyFromContext(ctx)
```
maraino: Related to the previous comment, just use a configuration struct or variadic methods when opening a TPM.

tpm/key.go
```diff
@@ -167,7 +180,7 @@ type AttestKeyConfig struct {
 // a random 10 character name is generated. If a Key with the same name exists,
 // `ErrExists` is returned. The Key won't be attested by an AK.
 func (t *TPM) CreateKey(ctx context.Context, name string, config CreateKeyConfig) (key *Key, err error) {
-	if err = t.open(goTPMCall(ctx)); err != nil {
+	if err = t.open(withMachineKey(goTPMCall(ctx), config.MachineKey)); err != nil {
```
maraino: gTPMCall here can also Go into the structure or variadic methods.

maraino (changes requested):
Make sure to use a release version of go-attestation once merged.

tpm/ak.go
joshdrake: This was done specifically to _avoid_ have a per-instance configuration setting for the TPM session, see: Per-key, not per-TPM-instance. [Herman's feedback on the original PR this is based on](https://github.com/smallstep/crypto/pull/1006#discussion_r3152660600)

hslatman:
> Make sure to use a release version of go-attestation once merged.

So far we've always used a commit. We didn't release from `surrogate`, because we always wanted to upstream it. There were some erroneous tags while you were on vacation, but I deleted those again.

tpm/ak.go
joshdrake: 4e76861165bc970908bef7f3b775dfa805808f6e

tpm/tpm.go
joshdrake: 4e76861165bc970908bef7f3b775dfa805808f6e

tpm/key.go
joshdrake: 4e76861165bc970908bef7f3b775dfa805808f6e

tpm/caps.go
```diff
@@ -42,7 +42,7 @@ func (t *TPM) GetCapabilities(ctx context.Context) (caps *Capabilities, err erro
 		return t.caps, nil
 	}
 
-	if err = t.open(goTPMCall(ctx)); err != nil {
+	if err = t.open(goTPMCall(ctx), openOptions{}); err != nil {
```
maraino: `goTPMCall` should be in `openOptions`

tpm/key.go
```diff
@@ -71,7 +72,7 @@ func (k *Key) Public() crypto.PublicKey {
 		err error
 		ctx = context.Background()
 	)
-	if err = k.tpm.open(ctx); err != nil {
+	if err = k.tpm.open(ctx, openOptions{k.machineKey}); err != nil {
```
maraino:
```suggestion
	if err = k.tpm.open(ctx, openOptions{machineKey: k.machineKey}); err != nil {
```

tpm/tpm.go
```diff
@@ -160,10 +160,14 @@ func New(opts ...NewTPMOption) (*TPM, error) {
 	}, nil
 }
 
+type openOptions struct {
+	machineKey bool
+}
```
maraino, line 169: Add the goTPMCall option.

maraino (changes requested):
Add `goTPMCall` as an option.

tpm/tpm.go
joshdrake: [fef78a2](https://github.com/smallstep/crypto/pull/1021/commits/fef78a254eb332b9552da3225264cbac6810f812)

tpm/caps.go
joshdrake: [fef78a2](https://github.com/smallstep/crypto/pull/1021/commits/fef78a254eb332b9552da3225264cbac6810f812)

tpm/key.go
joshdrake: [fef78a2](https://github.com/smallstep/crypto/pull/1021/commits/fef78a254eb332b9552da3225264cbac6810f812)

tpm/errors.go
```diff
@@ -15,3 +16,26 @@ var ErrExists = errors.New("already exists")
...
+// in-TPM (NCrypt PCP / attest) deletion failed but the file-store
+// entry was successfully removed. Callers can use [errors.As] to
+// detect this case and decide whether to treat it as a transient
+// failure (the named entry is gone from local bookkeeping; the
+// underlying TPM key may or may not still exist).
+//
+// Cleaning up the file-store entry even on PCP failure prevents
+// repeated retries from re-encountering the same failing entry,
+// which is the failure mode that motivated this type.
+type PartialDeleteError struct {
```
maraino, line 30: If you want to use something like `errors.Is(err, &PartialDeleteError{})`, as it is it will return always false. However this will work and return true:

```go
var partialDeleteErr *PartialDeleteError
errors.As(err, &partialDeleteErr)
```

If you want to make it work with `errors.Is()` you need to implement something like this:
```go
func (e *PartialDeleteError) Is(target error) bool {
	_, ok := target.(*PartialDeleteError)
	return ok
}
```

maraino (commented):
The code looks good, but I have a couple of questions, one is related to the use of `errors.Is()` and you can see a comment about that. 

The second is if there might be any issues getting, listing or deleting AK or keys that `MachineKey`

tpm/errors.go
joshdrake: I wasn't really doing this for value checking, more for unwrapping with `errors.As` since the the struct carries the `Underlying` error and key `Name`.  The idea being that `DeleteKey` callers can unwrap `PartialDeleteError` and then take additional action based on the underlying error (eg: delete the PCPKSP file handle, maybe).

joshdrake:
> The second is if there might be any issues getting, listing or deleting AK or keys that `MachineKey`

by default only administrators or LocalSystem can perform key operations, but otherwise, no

maraino (approved):

PR #1030, opened 2026-05-18
Delete CNG private keys when CleanupCredentials removes expired certs

darkfronza:
CleanupCredentials previously deleted the certificate from the Windows store but left the paired CNG private key in the provider, accumulating orphan .PCPKSP blobs on disk as certificates were renewed.

- Add a `delete-key` URI flag on CAPIKMS.DeleteCertificate. When set, the cert's CNG key is removed via nCryptDeleteKey before the cert context is deleted, so a partial failure leaves the cert in place for retry rather than orphaning the key.
- Rewrite CAPIKMS.CleanupCredentials to walk the store directly and delete each expired cert together with its CNG key. Enumeration is restarted after every delete because CertDeleteCertificateFromStore frees the context the next find call would chain from.
- TPMKMS.deleteCertificateFromWindowsCertificateStore now sets delete-key=true on its URI, since TPMKMS-managed certs have a 1:1 CNG key with no independent use.
- Update TestKMS_CleanupCredentials_capi to assert that both the certificate and the CNG private key are removed.

EFF-232

Change-Type: feature
Release-Note: yes
Audience: developer
Impact: low
Breaking: false


maraino (commented):
Capi already supports the DeleteKey method, why don't use just that in TPMKMS?

kms/tpmkms/tpmkms.go
```diff
@@ -1095,6 +1095,10 @@ func (k *TPMKMS) deleteCertificateFromWindowsCertificateStore(req *apiv1.DeleteC
 	uv := url.Values{}
 	uv.Set("store-location", location)
 	uv.Set("store", store)
+	// Also remove the CNG private key paired with the certificate; TPMKMS-managed
+	// certs have a 1:1 CNG key with no independent use, so leaving the .PCPKSP
+	// blob behind would orphan it on disk.
+	uv.Set("delete-key", "true")
```
hslatman: Is it clear that all call sites of `DeleteCertificate` are OK with also removing the key at all times?

Doing this by default, in `DeleteCertificate`, could be surprising. A key without a certificate is OK in general; maybe not that useful, but generally not an issue. The non-Windows certificate store path keeps the key stored on disk, so this makes it behave different in certain cases.

maraino: Agree, when we renew a certificate we might want to reuse the key.

darkfronza: Good call — addressed in 0e5a1a8.

`DeleteCertificate` no longer removes the key by default. The key is only deleted when the caller explicitly sets `delete-key=true` on the request URI; otherwise the certificate is removed and the CNG/PCP key is left in place, matching the non-Windows certificate-store behavior. So a renewal that wants to reuse the key just deletes the old cert without the flag.

Key deletion is now driven through `CleanupCredentials` instead (also opt-in via `delete-key` on its URI). When requested there, it first tries a full teardown via `TPMKMS.DeleteKey` — which removes both the TPM storage entry and the CNG/NCrypt container — and only falls back to deleting the container directly when `DeleteKey` cannot act on the key (storage not loadable, key not found, or TPM not openable).

kms/capi/capi.go
```diff
@@ -1188,48 +1181,137 @@ func (k *CAPIKMS) DeleteCertificate(req *apiv1.DeleteCertificateRequest) error {
...
+	var prevCert *windows.CertContext
+	for {
+		certHandle, err := findCertificateInStore(st,
+			encodingX509ASN|encodingPKCS7,
+			0,
+			findIssuerStr,
+			uintptr(unsafe.Pointer(wide(issuer))), prevCert)
+		if err != nil {
+			return false, fmt.Errorf("findCertificateInStore failed: %w", err)
+		}
```
maraino, line 1298: A comment saying that if the certificate does not exists this return nil, nil would be helpful so we don't need to navigate into the code to figure it out. 

It would be nicer if we return an `apiv1.NotFoundError` and then use `!errors.Is(err, apvi1.NotFoundError{})` to not fail in that case.

kms/capi/capi.go
```diff
@@ -1188,48 +1181,137 @@ func (k *CAPIKMS) DeleteCertificate(req *apiv1.DeleteCertificateRequest) error {
...
+		if certHandle == nil {
+			// prevCert was freed by the last findCertificateInStore call.
+			return false, nil
+		}
+
+		x509Cert, err := certContextToX509(certHandle)
+		if err != nil {
+			windows.CertFreeCertificateContext(certHandle)
```
maraino, line 1310: Do we need to always free the handle before returning? If we do, we are not doing it a few lines before.

And what about prevCert, it's also a handle, do we need to free it on every iteration?

kms/platform/kms_darwin_test.go
```diff
@@ -91,7 +91,7 @@ func TestKMS_CleanupCredentials_mackms(t *testing.T) {
 		assertion assert.ErrorAssertionFunc
 	}{
 		{"not implemented", platformKMS, args{&apiv1.CleanupCredentialsRequest{
-			Issuer:     chain[0].Issuer.CommonName,
+			Name:       "mackms:issuer=" + chain[0].Issuer.CommonName,
```
maraino: We should use `kms` instead of `mackms`
```suggestion
			Name:       "kms:issuer=" + chain[0].Issuer.CommonName,
```

kms/platform/kms_test.go
```diff
@@ -1324,7 +1324,9 @@ func TestKMS_CleanupCredentials(t *testing.T) {
 		assertion assert.ErrorAssertionFunc
 	}{
 		{"not implemented", softKMS, args{&apiv1.CleanupCredentialsRequest{
-			Issuer:     chain[0].Issuer.CommonName,
+			Name: uri.New("softkms", url.Values{
```
maraino: ditto

```suggestion
			Name: uri.New("kms", url.Values{
```

kms/platform/kms_windows_test.go
```diff
@@ -527,13 +537,21 @@ func TestKMS_CleanupCredentials_capi(t *testing.T) {
 		assertion assert.ErrorAssertionFunc
 	}{
 		{"ok", capiKMS, args{&apiv1.CleanupCredentialsRequest{
-			Issuer:     chain[0].Issuer.CommonName,
+			Name: uri.New("capi", url.Values{
```
maraino: We could use `kms` instead of capi

```suggestion
			Name: uri.New("kms", url.Values{
```

kms/tpmkms/tpmkms.go
```diff
@@ -1021,20 +1023,187 @@ func (k *TPMKMS) storeCertificateChainToWindowsCertificateStore(req *apiv1.Store
...
 	if req == nil {
 		return errors.New("cleanupCredentialsRequest cannot be nil")
 	}
 
-	if k.usesWindowsCertificateStore() {
-		return k.windowsCertificateManager.CleanupCredentials(req)
+	if !k.usesWindowsCertificateStore() {
+		// currently this API is a no-op on non Windows platforms.
+		return nil
 	}
```
maraino, line 1066: On platform kms we return `apiv1.NotImplementedError{}` if we don't support it, e.g. macOS, we should be consistent here.

kms/tpmkms/tpmkms.go
```diff
@@ -1021,20 +1023,187 @@ func (k *TPMKMS) storeCertificateChainToWindowsCertificateStore(req *apiv1.Store
...
+	})
+}
+
+// shouldFallbackToDirectKeyDelete reports whether a [TPMKMS.DeleteKey] error
+// means the CNG key container was never touched — the TPM storage couldn't be
+// loaded, the key wasn't present in storage, or the TPM couldn't be opened — so
+// the caller must fall back to deleting the key container directly.
+func shouldFallbackToDirectKeyDelete(err error) bool {
+	var notFound apiv1.NotFoundError
+	if errors.As(err, &notFound) || errors.Is(err, tpm.ErrNotFound) {
```
maraino: You should be able to use `errors.Is` too.

maraino (changes requested):
Some questions and some suggestions

kms/capi/capi.go
darkfronza: Added a clarifying comment in `ac63a96` so the `(nil, nil)` = "no (further) matching certificate" case is explicit without having to read `findCertificateInStore`.

I held off on changing `findCertificateInStore` to return `apiv1.NotFoundError`: it's shared by 6 other call sites, and each one turns the `nil` into a *context-specific* message (e.g. `certificate with hash=%x not found`, `certificate with issuer=%q not found`). A generic `NotFoundError` from the helper would lose that context, and here in the cleanup loop the `nil` is really a normal loop-terminator (we've walked all matching certs) rather than an error. Happy to do the broader refactor as a follow-up if you'd still prefer it.

kms/capi/capi.go
darkfronza: Good question — I added comments in `ac63a96` to make the ownership explicit. The model is:

- `CertFindCertificateInStore` frees the context passed as `prevCert` on each call (Win32 takes ownership of it), so we must **not** free `prevCert` ourselves per iteration — doing so would double-free.
- The explicit free here is needed because we bail out of the loop *without* chaining `certHandle` into another `findCertificateInStore` call.
- The earlier returns don't free because there's nothing to free: on the `findCertificateInStore` error path and the `certHandle == nil` path the handle is `nil` (and the previous one was already freed by that last call). On the delete path, `deleteCertContextAndMaybeKey` consumes/frees the handle.

So no leak and no double-free; it was just under-documented.

kms/platform/kms_darwin_test.go
darkfronza: Done in `ac63a96`.

kms/platform/kms_test.go
darkfronza: Done in `ac63a96`.

kms/platform/kms_windows_test.go
darkfronza: Done in `ac63a96`. To make this work end-to-end I also updated `KMS.CleanupCredentials` to run `req.Name` through `transformToURI` like the other platform methods — it was the only one not translating the generic `kms` scheme to the backend scheme.

kms/tpmkms/tpmkms.go
darkfronza: Done in `ac63a96`. `CleanupCredentials` now returns `apiv1.NotImplementedError{}` when the TPMKMS isn't configured with the Windows certificate store, instead of silently returning `nil`. Updated `TestKMS_CleanupCredentials_other` accordingly (the TPMKMS-available path now expects `NotImplementedError` too).

kms/tpmkms/tpmkms.go
darkfronza: Done in `ac63a96` — switched to `errors.Is(err, apiv1.NotFoundError{})` (it has an `Is` method), dropping the `errors.As` + temp variable.

kms/tpmkms/tpmkms.go
hslatman: This is an intentional `nil` return, as using the Windows certificate store in `tpmkms` is optional based on provided configuration.

kms/capi/capi.go
maraino: What about in the error on line 1319? Is it free inside the call to `deleteCertContextAndMaybeKey`? 

And if it is and `deleteCertContextAndMaybeKey` doesn't fail, would the set of `prevCert = certHandle` be a problem for the next call in the loop?

kms/apiv1/options.go
```diff
@@ -158,6 +158,55 @@ func (e NotFoundError) Is(target error) bool {
...
+type KeyError struct {
+	// Name identifies the key that failed, usually a KMS URI.
+	Name string
+	// Err is the underlying failure.
+	Err error
+}
+
+func (e *KeyError) Error() string {
+	return fmt.Sprintf("key %q: %v", e.Name, e.Err)
+}
```
maraino: The message of this error will look weird, and might not tell us what is going on. I don't see the need for distinguishing this type, and it might make more sense of properly wrap each error. For example:

```go
signer, err := key.Signer(context.Background())
if err != nil {
	errs = append(errs, fmt.Errorf("failed to create a signer for %q: %w", uri, err)
	continue
}
```

Or just the error.

kms/capi/capi.go
darkfronza: `deleteCertContextAndMaybeKey` always free the cert context, even on failure.
We don't continue the loop here, take a look at the code:

```
if matchesSubject && x509Cert.NotAfter.Before(now) {
			// deleteCertContextAndMaybeKey consumes certHandle (the cert delete
			// frees it), so no explicit free is needed on this path.
			if err := deleteCertContextAndMaybeKey(certHandle, deleteKey); err != nil {
				return false, fmt.Errorf("failed deleting expired certificate (serial %s): %w", x509Cert.SerialNumber.Text(16), err)
			}
			return true, nil
}
// THIS IS ONLY executed when we don't have a cert match, thus skipped deleteCertContextAndMaybeKey
// Carry certHandle forward; the next findCertificateInStore call frees it.
prevCert = certHandle
```

kms/tpmkms/tpmkms.go
```diff
@@ -1021,20 +1038,227 @@ func (k *TPMKMS) storeCertificateChainToWindowsCertificateStore(req *apiv1.Store
...
+	var partial *apiv1.PartialError
+	switch {
+	case errors.As(err, &partial):
+		// keep resp and partial; fall through to build the map from whatever
+		// keys were successfully returned.
+	case err != nil:
+		return nil, err
+	}
+
+	errs := make([]error, 0)
```
maraino: More common, and `(errs == nil) == true`:
```suggestion
	var errs []error
```

kms/tpmkms/tpmkms.go
```diff
@@ -1375,70 +1612,125 @@ func (k *TPMKMS) SearchKeys(req *apiv1.SearchKeysRequest) (*apiv1.SearchKeysResp
...
+	resp := &apiv1.SearchKeysResponse{Results: results}
+	if len(errs) > 0 {
+		// Best-effort contract: the successfully loaded keys are returned in
+		// resp alongside a *apiv1.PartialError describing the keys that could
+		// not be loaded. Callers that require completeness can treat any error
+		// as fatal; callers that can make progress with a subset (such as
+		// CleanupCredentials) inspect it with errors.As and keep using resp.
+		return resp, &apiv1.PartialError{Errors: errs}
+	}
+	return resp, nil
```
maraino: A couple of things:

1. It might be possible that we return a partial error when results is empty.

2. What is the use case of PartialErrors here? If `key.Signer()` inside `buildSearchKeyResults` fail will be related to some kind of corruption on the TPM, probably PCPKEY-related.

kms/platform/kms.go
```diff
@@ -283,22 +283,44 @@ func (k *KMS) SearchKeys(req *apiv1.SearchKeysRequest) (*apiv1.SearchKeysRespons
...
 		r.Query = query
 		resp, err := km.SearchKeys(r)
-		if err != nil {
+
+		// A *apiv1.PartialError means the backend returned usable results
+		// alongside per-key failures; patch and forward those results, keeping
+		// the partial error so callers can still inspect it. Any other error is
+		// fatal.
+		var partial *apiv1.PartialError
+		if err != nil && !errors.As(err, &partial) {
```
maraino: An custom implementation of Is would be helpful.

maraino (changes requested):

kms/apiv1/options.go
darkfronza: Agreed — dropped the `KeyError` type in 69f6c16. The signer failure is now wrapped directly: `fmt.Errorf("failed to create a signer for %q: %w", uri, err)`, so the message reads sensibly and there's no extra type to carry.

kms/tpmkms/tpmkms.go
darkfronza: Done in 69f6c16 — using `var errs []error`.

kms/tpmkms/tpmkms.go
darkfronza: Addressed in 69f6c16 — dropped `PartialError` too. `SearchKeys` now returns the keys it could load plus `errors.Join` of the failures, no custom type.

On the use case: this is exactly the `NTE_BAD_KEYSET` scenario from the original report. A single corrupt PCP key made `key.Signer()` fail, which made `SearchKeys` fail atomically, which in turn made `CleanupCredentials` (it builds the SKI→key map up front when `delete-key=true`) bail before deleting *any* expired certificate. Best-effort enumeration lets the rest of the keys come back so cleanup can proceed; the corrupt key's cert just falls through to the direct CNG-deletion path.

On point 1 (partial error with empty results): yes, that's intended now. If every key fails to load, the caller gets an empty result set plus the joined error. `keyNamesBySubjectKeyID` then returns an empty map, and `CleanupCredentials` deletes each expired cert via the direct path — so an empty-but-errored result degrades gracefully rather than being a special case.

kms/platform/kms.go
darkfronza: No longer needed after 69f6c16 — with the typed error gone, the wrapper doesn't inspect the error at all. It just forwards the (patched) results alongside whatever error came back when the response is non-nil, so there's no `errors.As`/`Is` to implement here.

kms/tpmkms/tpmkms.go
hslatman: I'm OK with this change, given that the goal is to make the behavior consistent. Ideally we'd also support the operation when not using the Windows cert store, but that's not required in this PR.

kms/platform/kms.go
```diff
@@ -283,22 +283,41 @@ func (k *KMS) SearchKeys(req *apiv1.SearchKeysRequest) (*apiv1.SearchKeysRespons
 		r := clone(req)
 		r.Query = query
 		resp, err := km.SearchKeys(r)
-		if err != nil {
+
+		// SearchKeys is best-effort: it may return usable results alongside an
+		// error describing the keys it couldn't load. Forward whatever results
+		// came back (after patching their URIs) together with that error, so
+		// callers can use the subset and still see what failed. Only a response
+		// we can't patch is fatal.
```
maraino: Do we know why this could happen, it would be nice to document it.

maraino (dismissed):
LGTM, but it would be nice to document why SearchKeys might fail.

kms/platform/kms.go
darkfronza: Documented in 855a7d1. Expanded this comment to name the concrete causes (a key's signer failing to load from a corrupt/unreadable TPM or PCP/CNG keyset — e.g. NTE_BAD_KEYSET on Windows, or a CNG container removed out of band — or the TPM failing to enumerate), and added the same explanation to the `TPMKMS.SearchKeys` godoc.

maraino (approved):
lgtm

PR #1053, opened 2026-06-11
fix(windows): associate machine-scoped TPM keys with stored certificates

darkfronza:
## Summary

On Windows, certificates stored for machine-scoped TPM (Microsoft Platform Crypto Provider) keys were left **without an associated private key**, making them unusable (e.g. for Wi-Fi/EAP-TLS, VPN). This was surfaced by the step-agent's `windows-key-scope` work, which moves the agent's AK — and therefore its attested endpoint keys — to the local machine keyset.

### Root cause

The certificate→key association ran exclusively through `CryptFindCertificateKeyProvInfo` (discovery), which fails here for two compounding reasons:

1. **It is skipped.** `kms/platform`'s `transformToTPMKMS` sets `skip-find-certificate-key=true` on Windows by default, so discovery never runs for these certificates.
2. **It cannot find the key anyway.** `CryptFindCertificateKeyProvInfo` does not search the local machine keyset, so it cannot discover a machine-scoped PCP/TPM key even when asked.

Keeping endpoint keys user-scoped isn't an option: `tpm.AttestKey` requires an attested key to share its AK's scope (an attest-session constraint), and the AK is machine-scoped.

### Fix — associate explicitly instead of by discovery

- **tpmkms** hands the CAPI layer the exact key when storing to the Windows certificate store: its CNG container name (`tpm.ApplicationKeyName`, the `app-`-prefixed key name used to persist the key), the TPM provider, and whether the key lives in the local machine keyset.
- **capi** sets `CERT_KEY_PROV_INFO_PROP_ID` directly via `setCertificateKeyProvInfo` when a container name is supplied. This needs no container enumeration, so it never prompts for a smart card and runs even when `skip-find-certificate-key` is set; it marks the prov info with `CRYPT_MACHINE_KEYSET` for machine-scoped keys.
- The `CRYPT_KEY_PROV_INFO` struct is corrected to pointer/DWORD fields so it can be used to *set* the property (it was `int`-typed and effectively read-only).
- The discovery fallback (used when no key is named) now searches both the user and machine keysets for machine-location certificates.
- `tpm.ApplicationKeyName` is exported as the single source of truth for the persisted application-key name, shared with tpmkms.

## Testing

- `GOOS=windows go build ./kms/... ./tpm/...` and host `go build` both clean; `go vet` shows only a pre-existing `unsafe.Pointer` line; `tpm` name tests pass.
- Verified end-to-end on a Windows host with a TPM: a freshly-issued hardware-attested endpoint certificate now has its private key linked in `LocalMachine\My` (previously orphaned).

### DIFF
--- a/kms/capi/capi.go
+++ b/kms/capi/capi.go
@@ -54,6 +54,7 @@ const (
 	IssuerNameArg                = "issuer"
 	KeySpec                      = "key-spec"                  // 0, 1, 2; none/NONE, at_keyexchange/AT_KEYEXCHANGE, at_signature/AT_SIGNATURE
 	SkipFindCertificateKey       = "skip-find-certificate-key" // skips looking up certificate private key when storing a certificate
+	MachineKeysetArg             = "key-machine-keyset"        // when storing a certificate, associate it with a key in the local machine keyset
 )
 
 const (
@@ -84,6 +85,7 @@ var signatureAlgorithmMapping = map[apiv1.SignatureAlgorithm]string{
 
 type uriAttributes struct {
 	containerName             string
+	providerName              string
 	hash                      []byte
 	storeLocation             string
 	storeName                 string
@@ -97,6 +99,7 @@ type uriAttributes struct {
 	description               string
 	keySpec                   string
 	skipFindCertificateKey    bool
+	machineKeyset             bool
 	pin                       string
 }
 
@@ -127,6 +130,7 @@ func parseURI(rawuri string) (*uriAttributes, error) {
 
 	return &uriAttributes{
 		containerName:             u.Get(ContainerNameArg),
+		providerName:              u.Get(ProviderNameArg),
 		hash:                      hashValue,
 		storeLocation:             cmp.Or(u.Get(StoreLocationArg), UserStoreLocation),
 		storeName:                 cmp.Or(u.Get(StoreNameArg), MyStore),
@@ -140,6 +144,7 @@ func parseURI(rawuri string) (*uriAttributes, error) {
 		description:               u.Get(DescriptionArg),
 		keySpec:                   u.Get(KeySpec),
 		skipFindCertificateKey:    u.GetBool(SkipFindCertificateKey),
+		machineKeyset:             u.GetBool(MachineKeysetArg),
 		pin:                       u.Pin(),
 	}, nil
 }
@@ -914,13 +919,35 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
 	}
 	defer windows.CertFreeCertificateContext(certContext)
 
-	// looking up the certificate private key is performed by default, but is made optional,
-	// so that looking up the private key for e.g. intermediate certificates can be skipped.
-	// If not skipped, looking up a private key can prompt the user to insert/select a smart
-	// card, which is usually not what we want to happen.
-	if !u.skipFindCertificateKey {
+	// Associate the certificate with its private key.
+	switch {
+	case u.containerName != "":
+		// The exact key is known (the caller supplied its container name, and
+		// usually its provider). Associate it explicitly rather than by
+		// discovery. This is required for machine-scoped Microsoft Platform
+		// Crypto Provider (TPM) keys: CryptFindCertificateKeyProvInfo does not
+		// search the local machine keyset and so cannot find them. Explicit
+		// association does not enumerate containers and never prompts for a
+		// smart card, so it runs even when skip-find-certificate-key is set.
+		var flags uint32
+		if u.machineKeyset {
+			flags |= CRYPT_MACHINE_KEYSET
+		}
+		if err := setCertificateKeyProvInfo(certContext, u.containerName, u.providerName, flags, ncryptKeySpec); err != nil {
+			return fmt.Errorf("failed associating certificate with key %q: %w", u.containerName, err)
+		}
+	case !u.skipFindCertificateKey:
+		// No specific key was named, so fall back to discovery. Looking up the
+		// private key can prompt the user to insert/select a smart card, which
+		// is why it is skipped for e.g. intermediate certificates. When the
+		// certificate is stored in the machine location its key may live in the
+		// local machine keyset, so search both keysets.
+		var keysetFlags uint32
+		if u.storeLocation == MachineStoreLocation {
+			keysetFlags = CRYPT_FIND_MACHINE_KEYSET_FLAG | CRYPT_FIND_USER_KEYSET_FLAG
+		}
 		// TODO: not finding the associated private key is not a dealbreaker, but maybe a warning should be issued
-		cryptFindCertificateKeyProvInfo(certContext)
+		cryptFindCertificateKeyProvInfo(certContext, keysetFlags)
 	}
 
 	if u.friendlyName != "" {
@@ -969,6 +996,12 @@ func (k *CAPIKMS) StoreCertificateChain(req *apiv1.StoreCertificateChainRequest)
 			FriendlyNameArg:        []string{u.friendlyName},
 			DescriptionArg:         []string{u.description},
 			SkipFindCertificateKey: []string{strconv.FormatBool(u.skipFindCertificateKey)},
+			// Forward the key association hints so the leaf certificate is
+			// linked to its private key (see StoreCertificate). Intermediates
+			// below have no associated key and intentionally omit these.
+			ContainerNameArg: []string{u.containerName},
+			ProviderNameArg:  []string{u.providerName},
+			MachineKeysetArg: []string{strconv.FormatBool(u.machineKeyset)},
 		}).String(),
 		Certificate: leaf,
 	}); err != nil {
--- a/kms/capi/capi.go
+++ b/kms/capi/capi.go
@@ -113,6 +113,15 @@ func (u *uriAttributes) isMachineKeyset() bool {
 		(u.keyScope == "" && u.storeLocation == MachineStoreLocation)
 }
 
+// isUserKeyset reports whether the certificate's private key is explicitly
+// scoped to the current user's keyset. Unlike isMachineKeyset it deliberately
+// does not fall back to the store location: when "key-scope" is unset the keyset
+// is left unspecified so key discovery can search both containers, matching
+// CryptFindCertificateKeyProvInfo's historical default.
+func (u *uriAttributes) isUserKeyset() bool {
+	return u.keyScope == UserStoreLocation
+}
+
 func parseURI(rawuri string) (*uriAttributes, error) {
 	u, err := uri.ParseWithScheme(Scheme, rawuri)
 	if err != nil {
@@ -949,11 +958,16 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
 	case !u.skipFindCertificateKey:
 		// No specific key was named, so fall back to discovery. Looking up the
 		// private key can prompt the user to insert/select a smart card, which
-		// is why it is skipped for e.g. intermediate certificates. Restrict the
-		// search to the keyset that matches the key scope (machine keys live in
-		// the machine keyset, which the default user-only search would miss).
-		keysetFlags := CRYPT_FIND_USER_KEYSET_FLAG
-		if u.isMachineKeyset() {
+		// is why it is skipped for e.g. intermediate certificates. With no
+		// keyset flag CryptFindCertificateKeyProvInfo searches both the user and
+		// machine containers (its documented default); restrict the search only
+		// when the key scope is known, so a machine-scoped key is found while
+		// the legacy both-containers behavior is preserved otherwise.
+		var keysetFlags uint32
+		switch {
+		case u.isUserKeyset():
+			keysetFlags = CRYPT_FIND_USER_KEYSET_FLAG
+		case u.isMachineKeyset():
 			keysetFlags = CRYPT_FIND_MACHINE_KEYSET_FLAG
 		}
 		// TODO: not finding the associated private key is not a dealbreaker, but maybe a warning should be issued
--- a/kms/capi/capi.go
+++ b/kms/capi/capi.go
@@ -54,7 +54,7 @@ const (
 	IssuerNameArg                = "issuer"
 	KeySpec                      = "key-spec"                  // 0, 1, 2; none/NONE, at_keyexchange/AT_KEYEXCHANGE, at_signature/AT_SIGNATURE
 	SkipFindCertificateKey       = "skip-find-certificate-key" // skips looking up certificate private key when storing a certificate
-	MachineKeysetArg             = "key-machine-keyset"        // when storing a certificate, associate it with a key in the local machine keyset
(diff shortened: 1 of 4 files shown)


kms/capi/capi.go
```diff
@@ -54,6 +54,7 @@ const (
 	IssuerNameArg                = "issuer"
 	KeySpec                      = "key-spec"                  // 0, 1, 2; none/NONE, at_keyexchange/AT_KEYEXCHANGE, at_signature/AT_SIGNATURE
 	SkipFindCertificateKey       = "skip-find-certificate-key" // skips looking up certificate private key when storing a certificate
+	MachineKeysetArg             = "key-machine-keyset"        // when storing a certificate, associate it with a key in the local machine keyset
```
maraino: This argument is always referenced as machine keyset, why are we using here `key-machine-keyset` instead of just `machine-keyset`?

```suggestion
	MachineKeysetArg             = "machine-keyset"            // when storing a certificate, associate it with a key in the local machine keyset
```

kms/capi/capi.go
```diff
@@ -914,13 +919,35 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
...
 	defer windows.CertFreeCertificateContext(certContext)
 
-	// looking up the certificate private key is performed by default, but is made optional,
-	// so that looking up the private key for e.g. intermediate certificates can be skipped.
-	// If not skipped, looking up a private key can prompt the user to insert/select a smart
-	// card, which is usually not what we want to happen.
-	if !u.skipFindCertificateKey {
+	// Associate the certificate with its private key.
+	switch {
+	case u.containerName != "":
```
maraino: Not related to this, but this attribute name is confusing, it is associated to `key`, I would name it `keyName`. And if you look docs for `CRYPT_FIND_MACHINE_KEYSET_FLAG` you will see [this](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptfindcertificatekeyprovinfo#parameters):

| Value | Meaning |
| - | - |
| CRYPT_FIND_USER_KEYSET_FLAG | Restricts the search to the user container. The default is to search both the user and machine containers.|
| CRYPT_FIND_MACHINE_KEYSET_FLAG | Restricts the search to the machine container. The default is to search both the user and machine containers. |
| CRYPT_FIND_SILENT_KEYSET_FLAG | The application requests that the CSP not display any user interface (UI) for this context. If the CSP must display the UI to operate, the call fails and the NTE_SILENT_CONTEXT error code is set as the last error. | 

It mentions for the "user" and "machine" container. 

I am working if instead of using `key-machine-keyset=true` (or `machine-keyset=true`) does it make more sense to use `container=machine` (or `container=user`) to restrict the search on the user container.

kms/capi/capi.go
maraino: See next comment about containers.

kms/capi/capi.go
```diff
@@ -914,13 +919,35 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
...
+		// association does not enumerate containers and never prompts for a
+		// smart card, so it runs even when skip-find-certificate-key is set.
+		var flags uint32
+		if u.machineKeyset {
+			flags |= CRYPT_MACHINE_KEYSET
+		}
+		if err := setCertificateKeyProvInfo(certContext, u.containerName, u.providerName, flags, ncryptKeySpec); err != nil {
+			return fmt.Errorf("failed associating certificate with key %q: %w", u.containerName, err)
+		}
+	case !u.skipFindCertificateKey:
```
maraino, line 958: Note that platform KMS, sets this by default to avoid a popup asking to look for a key in a smart card, see https://github.com/smallstep/crypto/blob/444dcf9a7d709c42402b98252de1836b69244a71/kms/platform/kms_tpm.go#L64-L68

kms/capi/capi.go
```diff
@@ -914,13 +919,35 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
...
+		// No specific key was named, so fall back to discovery. Looking up the
+		// private key can prompt the user to insert/select a smart card, which
+		// is why it is skipped for e.g. intermediate certificates. When the
+		// certificate is stored in the machine location its key may live in the
+		// local machine keyset, so search both keysets.
+		var keysetFlags uint32
+		if u.storeLocation == MachineStoreLocation {
+			keysetFlags = CRYPT_FIND_MACHINE_KEYSET_FLAG | CRYPT_FIND_USER_KEYSET_FLAG
+		}
```
maraino: I'm not sure if it's windows that is confusing, but we have `user` and `machine` store locations, and according to the above docs, `user` and `machine` containers. Are both two different things? 

But if they are the same thing is weird to use `store-location=machine` and looks into the `user` container 🫤

kms/capi/capi.go
maraino: Looking at docs about [CRYPT_KEY_PROV_INFO](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-crypt_key_prov_info) they mention that `pwszContainerName` is the name of the "key container", so a more appropriate name would be `u.keyContainerName`

kms/capi/ncrypt_windows.go
```diff
@@ -587,6 +611,51 @@ func cryptFindCertificateKeyProvInfo(certContext *windows.CertContext) error {
...
+		dwKeySpec:         dwKeySpec,
+	}
+
+	// CertSetCertificateContextProperty copies the structure and the strings it
+	// references into the certificate context, so the Go-allocated memory only
+	// needs to stay alive for the duration of the call.
+	err = certSetCertificateContextProperty(certContext, CERT_KEY_PROV_INFO_PROP_ID, uintptr(unsafe.Pointer(&info)))
+	runtime.KeepAlive(container)
+	runtime.KeepAlive(provider)
+	runtime.KeepAlive(&info)
```
maraino: I'm surprised this is required.

kms/tpmkms/tpmkms.go
```diff
@@ -989,16 +990,31 @@ func (k *TPMKMS) storeCertificateChainToWindowsCertificateStore(req *apiv1.Store
...
+		"friendly-name":               []string{o.friendlyName},
+		"description":                 []string{o.description},
+		"skip-find-certificate-key":   []string{skipFindCertificateKey},
+		"intermediate-store-location": []string{intermediateCAStoreLocation},
+		"intermediate-store":          []string{intermediateCAStore},
+	}
+	if o.name != "" {
+		v.Set("key", tpm.ApplicationKeyName(o.name))
+		v.Set("provider", microsoftPCP)
+		v.Set("key-machine-keyset", strconv.FormatBool(o.isMachineKey()))
```
maraino: So this is always true if `key-scope=machine` or `store-location=machine`. Let's use this logic in capi instead of adding a new parameter.

maraino (changes requested):
I would try to avoid the new parameter, we know using tpmkms parameter what would be the value of the `key-machine-keyset`, let's use the same logic, adding the already existing `key-scope` if needed.

I also have some comments about Microsoft naming.

kms/capi/capi.go
maraino: See last comment, perhaps we can avoid this parameter reusing tpmkms ones.

kms/capi/capi.go
darkfronza: There are two scopes, CurrentUser and LocalMachine, and both apply to either certificates or keys.
Every Principal on Windows has it's own, isolated, CurrentUser scope for both keys and certs.
LocalMachine is as it name says, a machine scope, shared by all users.
I think perhaps store-location should be renamed to cert-store-location.
Yeah I must fix this one, if store-locaction=machine should not lookup key in user scope, I'll fix this one.

kms/capi/capi.go
darkfronza: Done in b0e2839 — dropped the `key-machine-keyset` parameter entirely. tpmkms now forwards the existing `key-scope`, and capi resolves the keyset itself via `uriAttributes.isMachineKeyset()`, mirroring `tpmkms.isMachineKey()` (`key-scope`, falling back to `store-location`). So no new parameter, and the resolution logic lives in capi as suggested.

kms/tpmkms/tpmkms.go
darkfronza: Done in b0e2839 — moved this resolution into capi. tpmkms forwards `key-scope`; capi's `isMachineKeyset()` computes `key-scope=="machine" || (key-scope=="" && store-location=="machine")`, the same logic, so no extra parameter is needed.

kms/capi/capi.go
darkfronza: Renamed the `uriAttributes.containerName` field to `keyContainerName` in b0e2839 (matching `pwszContainerName` / "key container"). Left the public `ContainerNameArg = "key"` constant and the separate `CAPISigner.containerName` field unchanged to keep the diff focused.

kms/capi/capi.go
darkfronza: Agreed — they're two independent scopes (CurrentUser vs LocalMachine), each applying to certs and keys. The code now models them separately: `store-location` is the certificate store, and the new `key-scope` carries the key's keyset. capi derives the keyset from `key-scope`, falling back to `store-location` only when `key-scope` is unset (back-compat). The discovery fallback now restricts to the matching keyset rather than probing both, so a machine cert no longer looks in the user container. Renaming `store-location` → `cert-store-location` would read better but it's a breaking change to a public arg, so I left it for a separate discussion.

kms/capi/capi.go
darkfronza: Right — and that default is exactly why discovery wasn't associating these certs. The fix doesn't rely on discovery: when a key container name is supplied, capi sets `CERT_KEY_PROV_INFO_PROP_ID` explicitly, which doesn't enumerate containers or prompt for a smart card, so it runs even with `skip-find-certificate-key=true`.

kms/capi/ncrypt_windows.go
darkfronza: Removed in b0e2839. `CertSetCertificateContextProperty` copies the struct and its referenced strings, and `info` keeps the only references reachable through the call, so `KeepAlive` isn't needed — this now matches the other property helpers (e.g. `cryptSetCertificateFriendlyName`).

darkfronza:
Thanks for the review. Pushed b0e2839 addressing the feedback:

- Dropped the new `key-machine-keyset` parameter; tpmkms forwards the existing `key-scope` and capi resolves the keyset itself (same logic as `tpmkms.isMachineKey()`, falling back to `store-location`).
- Discovery fallback now restricts to the matching keyset by scope instead of setting both flags.
- Renamed the `uriAttributes.containerName` field to `keyContainerName`.
- Removed the `runtime.KeepAlive` calls in `setCertificateKeyProvInfo`.

One thing worth a look: the discovery fallback now restricts to the scope's keyset both ways (machine→machine, user→user), whereas before it searched both. It's more precise and matches the "a machine cert shouldn't probe the user container" point, but it's a subtle behavior change for the general (non-tpmkms) `StoreCertificate` discovery path. Since the tpmkms path now uses explicit association and rarely hits discovery, the risk is low — happy to revert just that part to the both-keysets default if you'd prefer.

The fix is verified working end-to-end on a Windows TPM host (attested endpoint cert now has its private key linked in `LocalMachine\My`).

kms/capi/capi.go
```diff
@@ -914,13 +929,35 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
...
+	case !u.skipFindCertificateKey:
+		// No specific key was named, so fall back to discovery. Looking up the
+		// private key can prompt the user to insert/select a smart card, which
+		// is why it is skipped for e.g. intermediate certificates. Restrict the
+		// search to the keyset that matches the key scope (machine keys live in
+		// the machine keyset, which the default user-only search would miss).
+		keysetFlags := CRYPT_FIND_USER_KEYSET_FLAG
+		if u.isMachineKeyset() {
+			keysetFlags = CRYPT_FIND_MACHINE_KEYSET_FLAG
+		}
```
maraino: It can be ok, but I'm not sure if it will be better to use both by default, I think leaving the keysetFlags to 0 will do it.

maraino (changes requested):
Looks good, but I think we might leave keysetFlags as 0 by default if `u.isMachineKeySet()` is false. See comment.

kms/capi/capi.go
maraino: For reference see https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptfindcertificatekeyprovinfo

darkfronza: Updated on https://github.com/smallstep/crypto/pull/1053/commits/25d69a9d26f196eb1267a0e46eb4cbdb41e266ad

maraino (approved):
lgtm

darkfronza:
Pushed 25d69a9 to make the key-discovery fallback backward-compatible.

**Context:** `CryptFindCertificateKeyProvInfo` searches **both** the user and machine key containers when no keyset flag is passed — that's its [documented default](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptfindcertificatekeyprovinfo) (*"The default is to search both the user and machine containers"*) and what this library did before the PR. My earlier revision defaulted the discovery path to `CRYPT_FIND_USER_KEYSET_FLAG`, which narrowed that to user-only — a behavior change for existing callers that don't specify a key scope.

**Change:** the discovery fallback now defaults to `0` (both containers) and only restricts when the key scope is actually known:

```go
var keysetFlags uint32 // 0 = search both containers (CryptFindCertificateKeyProvInfo default)
switch {
case u.isUserKeyset():    // explicit key-scope=user
    keysetFlags = CRYPT_FIND_USER_KEYSET_FLAG
case u.isMachineKeyset(): // key-scope=machine, or store-location=machine fallback
    keysetFlags = CRYPT_FIND_MACHINE_KEYSET_FLAG
}
```

Behavior:

| key-scope | store-location | search | vs. pre-PR |
| --- | --- | --- | --- |
| _(unset)_ | user (default) | both | preserved |
| _(unset)_ | machine | machine | narrowed (intended) |
| user | any | user | narrowed |
| machine | any | machine | narrowed |

So the common legacy case (no `key-scope`, default store) keeps the both-containers search, and we only narrow on an explicit signal.

**One deliberate asymmetry:** the new `isUserKeyset()` checks *only* explicit `key-scope`, while `isMachineKeyset()` keeps its `store-location=machine` fallback. That's intentional — adding a user-side fallback would make the defaulted `store-location=user` over-restrict to user-only and defeat the back-compat goal. The only edge it doesn't cover is an (exotic) machine-store cert whose key is in the *user* keyset, discovered without a container name; the explicit-association path — which the agent always takes — is unaffected.

For background, this is consistent with the canonical Windows behavior: a certificate's store location and its private key's keyset are independent axes (the link is the cert's `CERT_KEY_PROV_INFO_PROP_ID` / `CRYPT_KEY_PROV_INFO`, which has no store field), so it's correct to leave the keyset unspecified — and let discovery search both — unless the caller tells us the scope.

PR #1062, opened 2026-06-24
Use correct CNG-key value for 'dwKeySpec' when setting key provider info

joshdrake:
When the `dwProvType` member is `0` (which is the case for CNG keys), this value is passed as the `dwLegacyKeySpec` parameter to the `NCryptOpenKey` function. The correct `dwLegacyKeySpec` value for non-legacy CNG keys is `0`.

https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-crypt_key_prov_info

💔Thank you!

### DIFF
--- a/kms/capi/capi.go
+++ b/kms/capi/capi.go
@@ -952,7 +952,7 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
 		if u.isMachineKeyset() {
 			flags |= CRYPT_MACHINE_KEYSET
 		}
-		if err := setCertificateKeyProvInfo(certContext, u.keyContainerName, u.providerName, flags, ncryptKeySpec); err != nil {
+		if err := setCertificateKeyProvInfo(certContext, u.keyContainerName, u.providerName, flags, LEGACY_KEY_SPEC_NONE); err != nil {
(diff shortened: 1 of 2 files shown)


darkfronza (dismissed):

hslatman (dismissed):

kms/capi/capi.go
```diff
@@ -952,7 +952,7 @@ func (k *CAPIKMS) StoreCertificate(req *apiv1.StoreCertificateRequest) error {
 		if u.isMachineKeyset() {
 			flags |= CRYPT_MACHINE_KEYSET
 		}
-		if err := setCertificateKeyProvInfo(certContext, u.keyContainerName, u.providerName, flags, ncryptKeySpec); err != nil {
+		if err := setCertificateKeyProvInfo(certContext, u.keyContainerName, u.providerName, flags, LEGACY_KEY_SPEC_NONE); err != nil {
```
hslatman, line 955: Minor, but the comment on `setCertificateKeyProvInfo` refers to `CERT_NCRYPT_KEY_SPEC`. That could be updated to refer to `LEGACY_KEY_SPEC_NONE`.

joshdrake: 520ec32

hslatman (approved):

PR #1070, opened 2026-07-07
Make TPM dirstore cache size configurable

darkfronza:
## Summary

`NewDirstore` hardcoded the diskv in-memory read cache at 1 MiB. Because a store handle serves a key/AK blob from that cache after its first read, a handle keeps returning a blob even after a *separate* handle deletes it from the backing directory. Callers that mutate the directory out of band therefore have to rebuild their store handle to observe the change.

This adds a way to disable (or size) that cache.

## Changes

- **`tpm/storage`**: `NewDirstore(directory string, opts ...DirstoreOption)` is now variadic (backward compatible; existing callers keep the 1 MiB default). New `WithCacheSize(bytes uint64)` option; `0` disables caching so every read reflects current on-disk state.
- **`kms/tpmkms`**: thread the setting through the `storage-cache-size` URI parameter, parsed in both `ParseTPMOptions` and `New`. `New` now parses the URI before constructing the default store, so the setting is honored regardless of whether the storage directory comes from the URI or `apiv1.Options.StorageDirectory`.

## Motivation

The smallstep agent hit this on Windows: it recovers from an "orphaned AK" (blob present, backing PCP/CNG key deleted) by deleting the stale blob and letting the KMS recreate the AK. Its probe TPM handle had already cached the dead blob, forcing a handle rebuild after deletion. With `storage-cache-size=0` the probe handle reflects the deletion directly and the rebuild goes away. (Loading a key into the TPM dwarfs a blob read from disk, so caching that read buys little here.)

## Testing

- `TestNewDirstore_cacheSize` — default / custom / disabled map to the right `CacheSizeMax`.
- `TestDirstore_noCacheReflectsOutOfBandDelete` — a cached store serves a stale blob after an out-of-band delete; an uncached store reflects the delete.
- `TestParseDirstoreOptions` — URIs (`nil`, absent, `=0`, `=4096`, negative) thread through to the expected cache behavior via the public store API.

### DIFF
--- a/kms/tpmkms/tpmkms.go
+++ b/kms/tpmkms/tpmkms.go
@@ -190,36 +190,40 @@ func WithWindowsIntermediateStore(store, location string) Option {
 // ParseTPMOptions is a helper method that returns a slice of [tpm.NewTPMOption]
 // for the given URI.
 func ParseTPMOptions(u *uri.URI) []tpm.NewTPMOption {
-	var opts []tpm.NewTPMOption
-	if device := u.Get("device"); device != "" {
-		opts = append(opts, tpm.WithDeviceName(device))
-	}
-	if storageDirectory := u.Get("storage-directory"); storageDirectory != "" {
-		opts = append(opts, tpm.WithStore(storage.NewDirstore(storageDirectory, parseDirstoreOptions(u)...)))
-	}
-	return opts
+	return tpmOptionsForURI(u, "")
 }
 
-// newDirstore constructs the dirstore backing a [New] TPMKMS. It is a package
-// var so tests can observe the storage directory and options New resolves.
+// newDirstore constructs the dirstore backing the TPM. It is a package var so
+// tests can observe the storage directory and options it is built with.
 var newDirstore = storage.NewDirstore
 
-// resolveStorageDirectory returns the TPM storage directory and dirstore
-// options that [New] uses for the given options and parsed URI. The directory
-// precedence is: a storage-directory in the URI wins, otherwise
-// opts.StorageDirectory, otherwise the default "tpm". The dirstore options
-// (e.g. the cache size) come from the URI via parseDirstoreOptions.
-func resolveStorageDirectory(opts apiv1.Options, u *uri.URI) (string, []storage.DirstoreOption) {
-	directory := "tpm" // store TPM objects in a relative tpm directory by default.
-	if opts.StorageDirectory != "" {
-		directory = opts.StorageDirectory
-	}
+// tpmOptionsForURI returns the [tpm.NewTPMOption]s encoded in u — the device
+// name and a dirstore (with any storage-cache-size applied). It is the single
+// place that turns a URI into TPM options, shared by [ParseTPMOptions] and
+// [New].
+//
+// The store is built from the URI's storage-directory when present, otherwise
+// from defaultStorageDirectory; when both are empty no store option is added,
+// so the caller falls back to tpm.New's default store. The dirstore is created
+// through newDirstore so tests can observe it.
+func tpmOptionsForURI(u *uri.URI, defaultStorageDirectory string) []tpm.NewTPMOption {
+	var device, directory string
 	if u != nil {
-		if d := u.Get("storage-directory"); d != "" {
-			directory = d
-		}
+		device = u.Get("device")
+		directory = u.Get("storage-directory")
 	}
-	return directory, parseDirstoreOptions(u)
+	if directory == "" {
+		directory = defaultStorageDirectory
+	}
+
+	var opts []tpm.NewTPMOption
+	if device != "" {
+		opts = append(opts, tpm.WithDeviceName(device))
+	}
+	if directory != "" {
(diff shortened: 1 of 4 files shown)


kms/tpmkms/tpmkms_test.go
```diff
@@ -12,10 +12,61 @@ import (
...
+		uri  string
+		// reflectsDelete is true when the resulting store has caching disabled,
+		// so a blob deleted by another handle is observed on the next read.
+		reflectsDelete bool
+	}{
+		{"nil-uri", "", false},
+		{"no-cache-param", "tpmkms:storage-directory=x", false},
+		{"cache-disabled", "tpmkms:storage-cache-size=0", true},
+		{"cache-custom", "tpmkms:storage-cache-size=4096", false},
+		{"negative-ignored", "tpmkms:storage-cache-size=-1", false},
```
hslatman: Maybe change this so that the negative size becomes 0, and thus disabled. I believe that'll be less surprising.

darkfronza: Good call — done in ea21c43. A negative `storage-cache-size` now clamps to 0 (cache disabled) instead of being ignored, and the `negative-disables` test case asserts the disabled behavior.

kms/tpmkms/tpmkms.go
```diff
@@ -371,24 +393,29 @@ const (
...
+	var u *uri.URI
+
+	// Parse the URI up front so the default store below can honor any storage
+	// options it carries (e.g. storage-cache-size).
+	if opts.URI != "" {
+		if u, err = uri.ParseWithScheme(Scheme, opts.URI); err != nil {
+			return nil, fmt.Errorf("failed parsing %q as URI: %w", opts.URI, err)
+		}
+		uriOptions = ParseOptions(u)
+	}
```
hslatman: Could you add a case or two to verify that the correct storage directory and cache size ends up being used when `New` is called?

darkfronza: Added in 9f13682. I pulled the directory + cache resolution `New` performs into `resolveStorageDirectory` (URI `storage-directory` > `opts.StorageDirectory` > default `tpm`) — which also drops the earlier "two `WithStore`, last wins" construction — and added `TestResolveStorageDirectory` covering the directory precedence and that `storage-cache-size` threads through. Since `New` now delegates to that helper, the cases exercise its actual resolution logic.

darkfronza: Follow-up in f6cff88: the previous test only exercised the helper. Added `TestNew_storeConstruction`, which drives `New` itself and asserts the storage directory and the exact `CacheSize()` the backing dirstore is built with (0 and 4096 cases, plus default-enabled), via a `newDirstore` test seam. Also added `storage.Dirstore.CacheSize()` so the cache budget is observable.

kms/tpmkms/tpmkms.go
```diff
@@ -195,7 +195,51 @@ func ParseTPMOptions(u *uri.URI) []tpm.NewTPMOption {
 		opts = append(opts, tpm.WithDeviceName(device))
 	}
 	if storageDirectory := u.Get("storage-directory"); storageDirectory != "" {
-		opts = append(opts, tpm.WithStore(storage.NewDirstore(storageDirectory)))
+		opts = append(opts, tpm.WithStore(storage.NewDirstore(storageDirectory, parseDirstoreOptions(u)...)))
```
hslatman:
```suggestion
		opts = append(opts, tpm.WithStore(newDirstore(storageDirectory, parseDirstoreOptions(u)...)))
```

kms/tpmkms/tpmkms.go
```diff
@@ -371,26 +415,27 @@ const (
...
 
+	storageDirectory, dirstoreOptions := resolveStorageDirectory(opts, u)
+	tpmOpts := []tpm.NewTPMOption{
+		tpm.WithStore(newDirstore(storageDirectory, dirstoreOptions...)),
+	}
+	if u != nil {
+		if device := u.Get("device"); device != "" {
+			tpmOpts = append(tpmOpts, tpm.WithDeviceName(device))
+		}
+	}
```
hslatman, line 437: This now seems to replicate the logic from `ParseTPMOptions`. Not sure if the cache size can be incorporated there instead, so that the options are handled in a single place?

darkfronza: Good call — consolidated in eec0a45. `New` no longer replicates `ParseTPMOptions`: both now delegate to a single `tpmOptionsForURI(u, defaultStorageDirectory)` that builds the device + store (with `storage-cache-size` applied) in one place. `ParseTPMOptions` passes an empty default; `New` passes its resolved storage directory (opts.StorageDirectory, else `tpm`), and a URI `storage-directory` still wins. Removed `resolveStorageDirectory` accordingly.

kms/tpmkms/tpmkms.go
darkfronza: Addressed by the same consolidation (eec0a45): store construction now lives solely in `tpmOptionsForURI`, which uses the `newDirstore` seam — so this line went through that helper rather than a direct `storage.NewDirstore` call.

hslatman (approved):

PR #1072, opened 2026-07-16
Fix issues on gotip

hslatman:
This PR fixes two issues when I ran it against `gotip`:

- Go 1.27 adds [CertificateRequest.RawSignatureAlgorithm](https://go.dev/pkg/crypto/x509#CertificateRequest.RawSignatureAlgorithm), and property equality tests were failing on that
- A digest with wrong hash length was passed to sign by an ECDSA key. Its `Sign` method now checks if the length is correct, and returns an error otherwise.


maraino (approved):
lgtm, an improvement to this would be to clear all `Raw*` data that we need for the test to pass.

PR #1098, opened 2026-08-05
Add support for ML-DSA algorithms

maraino:
This commit adds support for creating keys and signing using ML-DSA-44, ML-DSA-65, and ML-DSA-87 algorithms on GCP cloud KMS, AWS KMS, softkms, and keyutil and pemutil packages.

It also upgrades Go to 1.26+, uses `go fix` to modernize Go, and changes the deprecated fields in ECDSA keys to use mainly the `ecdsa.PublicKey.Bytes()`, `ecdsa.PrivateKey.Bytes()` and `ecdsa.ParseUncompressedPublicKey()` methods.

We could use https://github.com/smallstep/step-kms-plugin/pull/364 to test other implementations.


kms/cloudkms/cloudkms_other.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"crypto/x509"
+
+	"cloud.google.com/go/kms/apiv1/kmspb"
+	"go.step.sm/crypto/kms/apiv1"
+)
+
+func patchSignatureAlgorithmMapping(m map[apiv1.SignatureAlgorithm]interface{}) map[apiv1.SignatureAlgorithm]interface{} {
+	return m
```
hslatman:
```suggestion
func patchSignatureAlgorithmMapping(m map[apiv1.SignatureAlgorithm]any) map[apiv1.SignatureAlgorithm]any {
	return m
```

kms/awskms/awskms_other.go
```diff
@@ -0,0 +1,9 @@
+//go:build !go1.27
+
+package awskms
+
+import "go.step.sm/crypto/kms/apiv1"
+
+func patchSignatureAlgorithmMapping(m map[apiv1.SignatureAlgorithm]interface{}) map[apiv1.SignatureAlgorithm]interface{} {
+	return m
```
hslatman:
```suggestion
func patchSignatureAlgorithmMapping(m map[apiv1.SignatureAlgorithm]any) map[apiv1.SignatureAlgorithm]any {
	return m
```

keyutil/key.go
```diff
@@ -122,17 +119,20 @@ func GenerateSigner(kty, crv string, size int) (crypto.Signer, error) {
 		return generateRSAKey(size)
 	case "OKP":
 		return generateOKPKey(crv)
+	case "AKP":
```
hslatman, line 135: I believe it would be best if we would also support supplying `kty=ML-DSA-44` and such: `AKP` is not something that's generally known and comes from JOSE/COSE; it's not used in general crypto jargon. 

This is also in line with OpenSSL:

```console
  openssl genpkey -algorithm ML-DSA-44 -out key.pem
  openssl genpkey -algorithm ML-DSA-65 -out key.pem
  openssl genpkey -algorithm ML-DSA-87 -out key.pem
```

Besides that, I believe it's also better to not pass the actual key type as `crv`, since ML-DSA has nothing to do with curves.

Also note that based on the definition of `AKP` it could even support e.g. `ES256` and the likes in addition to `ML-DSA-44`, as `ES256` defines both the curve P-256 with SHA256 as algorithm. Similar for `Ed25519`, which is fully specified by just that string.

From [RFC 9964](https://datatracker.ietf.org/doc/html/rfc9964):

```text
This section specifies a generic cryptographic key structure for use 
with algorithms not limited to those registered in this document. The 
Algorithm Key Pair (AKP) type is used to express public and private 
keys for use with algorithms.
```

kms/cloudkms/cloudkms_other.go
maraino: Fixed with 44b50c6.

I've run `go fix -tags tpmsimualtor ./...`, `go1.26.8 fix -tags tpmsimualtor ./...` on Linux and `go fix ./...` on windows.

kms/awskms/awskms_other.go
maraino: Fixed with 44b50c6.

keyutil/key.go
maraino: With f2cf7e1, we allow `ML-DSA-44`, `ML-DSA-65`, `ML-DSA-87`, `Ed25519`, and `X25519` as `kty`. We will still allow JOSE AKP and OKP.

maraino: I've also clarified the comments on 3e3cd44

hslatman (approved):

keyutil/key.go
hslatman: Awesome 😄

PR #1130, opened 2026-09-10
Add AGENTS.md agent guidance

tashian:
#### Name of feature:

Adds a `AGENTS.md` with agent guidance for this repository. It covers the package map, the `kms` backend registry and per-backend cgo/platform build tags, TPM simulator testing, generated files, the Makefile test/lint targets, and the error-wrapping and testing conventions in use.

`.claude/CLAUDE.md` is a one-line `@../AGENTS.md` import so Claude Code loads the same file.

Tracks [EFF-787](https://linear.app/smallstep/issue/EFF-787/add-claudemd-agent-guidance-to-repos-that-lack-one).

https://claude.ai/code/session_01JuSZFSx1cTEnqY55oquacV


AGENTS.md
```diff
@@ -0,0 +1,149 @@
...
+
+Guidance for AI coding agents working in this repository. Claude Code loads it through the one-line `@../AGENTS.md` import in `.claude/CLAUDE.md`.
+
+## Overview
+
+`go.step.sm/crypto` is Smallstep's shared cryptography library (Apache-2.0). It is a
+library only, no binaries: X.509 and SSH certificate templating, key generation and
+PEM/JWK handling, a pluggable KMS abstraction (software, AWS, GCP, Azure, PKCS #11,
+YubiKey, ssh-agent, TPM, macOS Keychain, Windows CAPI), and TPM 2.0 attestation.
+[`step`](https://github.com/smallstep/cli) and [`step-ca`](https://github.com/smallstep/certificates)
```
maraino, line 11: We could mention [`step-kms-plugin`](https://github.com/smallstep/step-kms-plugin) too

AGENTS.md
```diff
@@ -0,0 +1,149 @@
...
+are the main consumers; changes here ship to them by bumping the dependency, so keep
+exported APIs backward compatible. `go.mod` requires Go 1.26; CI tests `stable` and
+`oldstable`.
```
maraino, line 14: I suppose Claude knows, but should we clarify what `stable` and `oldstable` means?

AGENTS.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Commands
+
+```bash
+make bootstrap      # install golangci-lint, govulncheck, gotestsum into $(go env GOPATH)/bin
```
maraino, line 19: We will be using go tools with https://github.com/smallstep/crypto/pull/1131

AGENTS.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+Single test / single package:
+
+```bash
+go test -run TestNewCertificate ./x509util/
+CGO_ENABLED=1 go test -tags tpmsimulator -run TestTPM_CreateAK ./tpm/
+```
+
+`make test` and `make race` need `gotestsum` on `PATH`; `make lint` needs network
```
maraino, line 37: We don't need gotestsum with go tools, we will need network.

AGENTS.md
```diff
@@ -0,0 +1,149 @@
...
+  `apiv1.NotFoundError` / `apiv1.AlreadyExistsError` are the sentinel types backends return.
+- **No logging framework**; this is a library. Do not add one.
+- **Tests**: `testify` (`assert`/`require`), table-driven `t.Run` loops, fixtures under
+  each package's `testdata/`. `gomock` is used only for the Azure Key Vault client.
+  `minica` and `internal/testutil` are the shared helpers.
+- **Templates**: `x509util` and `sshutil` render JSON via `text/template` with the
+  `internal/templates` FuncMap (sprig plus `toTime`, `formatTime`, etc.); the `fail`
+  function surfaces user-visible template errors. The `Default*Template` constants are
+  consumed verbatim by `step-ca` provisioners, so changing them is a behavior change.
+- **Imports**: `goimports -local go.step.sm/crypto`; lint rules come from the shared
```
maraino, line 144: This will use go tools

maraino (changes requested):
We will need some changes after the tools PR

AGENTS.md
tashian: Added step-kms-plugin alongside step and step-ca in the consumers sentence.

AGENTS.md
tashian: Clarified: stable is the newest Go release, oldstable the previous minor, and code must build on both.

AGENTS.md
tashian: Updated for #1131: the bootstrap line is gone (the target is now a no-op) and every command is described as running through `go tool` from the go.mod tool block.

AGENTS.md
tashian: Reworded: nothing to install, but the first `go tool` invocation downloads the tool modules over the network, and `make lint` always needs network for `.golangci.yml`.

AGENTS.md
tashian: Now reads `make fmt` (`go tool goimports -local go.step.sm/crypto`).

.claude/CLAUDE.md
```diff
@@ -0,0 +1 @@
+@../AGENTS.md
```
hslatman, line 1: This is no longer necessary: https://code.claude.com/docs/en/changelog#2-1-277.

tashian: @azazeal

azazeal: That's great news, but the industry has a tendency to take back good things, so leaving it as is is fine, just a small indirection that nobody needs to think of going forward.

PR #1134, opened 2026-09-15
Add support for creating signers with (just) `hash` on MacKMS

hslatman:
When `CreateSigner` was called without a `label`, but with `hash` set, it would fail indicating the label is required.

This commit changes that behavior to allow the label to be empty if the hash is set on `CreateSigner` and `GetPublicKey`.

Note the algorithm switch at https://github.com/smallstep/crypto/pull/1134/changes#diff-cf9c95370f587578dc8508e15dafd67ca3264d2fa044503c1a5e0522ede504fbR472. Using ECDSA seemed to be the right thing to do given the name of the key.


joshdrake (approved):

kms/mackms/mackms_test.go
```diff
@@ -462,7 +469,7 @@ func TestMacKMS_CreateSigner(t *testing.T) {
 	kms := &MacKMS{}
 	resp, err := kms.CreateKey(&apiv1.CreateKeyRequest{
 		Name:               "mackms:label=test-p256",
-		SignatureAlgorithm: apiv1.SHA256WithRSA,
+		SignatureAlgorithm: apiv1.ECDSAWithSHA256,
```
joshdrake: 👍

maraino (approved):
It is weird that DeleteKey requires a label, why is that?

hslatman:
> It is weird that DeleteKey requires a label, why is that?

I wanted to keep the behavior change limited to what we needed now. For `CreateKey` I think we'll always want to require the `label`. For `DeleteKey` I can see it working if only `hash` is set, but that'll require some more testing.

